Issue #35, opened 2015-03-12
Some 'navigation' for the different 'topics' of ETLoader is needed


ChaelKruip:
When looking at a testing ground, the user should be able to navigate to the 'load profiles' section and vice versa. This will become more relevant if we also support more testing-ground functionality such as 'business cases' and 'market models'.

Perhaps we could make different tabs for the different 'topics'?

ChaelKruip:
Re-assigning to @dennisschoenmakers so he can shine his light on this and find someone to make this happen.

dennisschoenmakers:
See #48.

PR #48, opened 2015-03-26
New Layout

### LINKED ISSUE #40: Proposal: testing ground navigation
This is a working document for the design of an interface that accommodates the newly developed functionality for testing ground calculations.

After talking to @AlexanderWirtz I feel that one intuitive option would be to use a side-bar menu similar to that of the (professional) ETM:
ivy_page_key

There are several advantages to this:

* A certain amount of 'homogeneity' with the ETM
* Familiarity with how such a menu works (if the user is familiar with the ETM)
* Re-usability of design and framework

dennisschoenmakers:
First step in adding more interface structure.

This will close #40 and #35.

I cleaned up, added Bootstrap, added a header with a menu where all the links for the different pages can be found and added two static pages:
* home page
* how to page (content by @ChaelKruip)

#### Still need to be done
- [ ] Elaborate a bit more on the descriptions on the home page
- [ ] Elaborate a bit more on the How to page
- [ ] Maybe add some images/illustrations

#### Far flung ideas
* I think we need to add user accounts, so we can register who has added what.

@antw: could you merge at your will?


antw:
> @antw: could you merge at your will?

Done. *Huge* improvement, thanks!

> I think we need to add user accounts, so we can register who has added what.

100% in agreement.

PR #88, opened 2015-05-01
Issues/new form flow

grdw:


antw:
Thanks for this, @grdw. Today is a public holiday in the UK so I took it off; I'll take a proper look at this first thing on Tuesday.

antw:
Hi Gerard,

Thanks for the pull request; I think the new workflow is a big improvement over what we had before, and I think the client will find it much easier to use.

I have a couple of comments:

* While you’re welcome to make as many commits as you want during a work-in-progress, before merging into master it’s best to try to squash new features into fewer commits. Ideally, you want one commit per *significant* change.

  The reason for this is that the Git log for the master branch should be a readable history of what has happened, and it quickly becomes unreadable if it records every small iteration of developing a new feature.

  I *don’t* mean that you have to squash the entire branch to a single commit – despite some people advocating this – although smaller pull requests like “[mail setup](https://github.com/quintel/etloader/pull/86)” or "[updating links](https://github.com/quintel/etloader/pull/80)" could be. In this case, perhaps you might have:

  * Add default installed technology with InstalledTechnology#template
  * Add new technology profile selection classes (differentiation).
  * Add new testing ground form.

  It’s also quite hard to comment on specific changes when you have to look through 63 commits to find the one with the change in question. :grin:

  My workflow is normally something like:

  * Rebase branch against master (to minimise the amount of merge commits).
  * Interactive rebase to reorder related commits to be "next" to each other in the timeline.
  * Interactive rebase again to squash the related commits together, and update or rewrite commit messages.

  If I feel like that's too much trouble (which it might be for this pull request), I'll `git reset` the entire branch, and use [`git add --patch`](http://nuclearsquid.com/writings/git-add/) to selectively create new commits.

  See also: [Squashing Github pull requests into a single commit](http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit)

* I’m finding that submitting the new testing ground form after clicking Testing Grounds -> Create submits an empty `technology_profiles` field. It works fine if I load testing_grounds/new directly. Likely this is caused by Turbolinks.

I have some other comments, but I’ll save those until we’ve dealt with – or discussed – the two issues above.

grdw:
- First issue is resolved (and I finally learned how to interactive rebase :dancers: )

> I’m finding that submitting the new testing ground form after clicking Testing Grounds -> Create submits an empty technology_profiles field. It works fine if I load testing_grounds/new directly. Likely this is caused by Turbolinks.

- The second issue is kind of vague. Do you mean when importing from ETM?

antw:
> First issue is resolved (and I finally learned how to interactive rebase)

Great! :+1:  It's such a useful tool, I don't know how I'd manage without it these days.

> The second issue is kind of vague. Do you mean when importing from ETM?

I can trigger the error by:

1. Visit `/testing_grounds`
2. Click "Testing Grounds" in the menu, then "Create new"
3. Enter a "Testing ground name" in the form
4. Hit submit

If I skip step 1 and 2, and instead directly visit `testing_grounds/new` (by typing it into the address bar), I can submit the form fine; that's why I'm guessing it's caused by Turbolinks. I'm not really sure why this would happen, though, as the `technologies` and `topology_attributes[graph]` parameter are both submitted back to the server, and there doesn't appear to be anything different to the `technology_profiles` field and ACE editor are set up.

I have posted the [backtrace and submitted form parameters](https://gist.github.com/antw/9924a93f7f9c656ff195). If you can't reproduce it, let me know and I'll investigate a bit more (in case a browser extension is responsible).

grdw:
I also managed to trigger the error. But only by clicking 'Create new' from the menu, not when browsing there. 

A thing I notice though is that the syntax highlighting from ACE is off, whenever I can't submit it properly. So my guess is, that it would have something to do  with ACE. 

Right now I fixed it by also updating the textarea directly. It isn't that nice of a fix, but considering time..

antw:
@grdw, Chael wants to be able to test this on staging tomorrow, so the any other comments I have can wait; they're not important. :smiley:

I think the only remaining show-stopper is that old testing grounds aren't migrated to your new `technology_profiles` column, which means they'll no longer work. Personally, I don't think we need separate `technologies` and `technology_profiles` (after creating the testing ground, `technology_profiles` is all we need). Maybe you or @ChaelKruip have reasons for wanting to store both, but that discussion can be had another day...

Will it be difficult to migrate the old testing grounds to use the new column? I think the contents of the old `technologies` column are in the same format as the new `technology_profiles`.

grdw:
> I think the only remaining show-stopper is that old testing grounds aren't migrated to your new  `technology_profiles` column, which means they'll no longer work. 

I'll write a migration for this. I don't think it is going to be that hard.

> Personally, I don't think we need separate `technologies` and `technology_profiles`

I haven't really discussed this in detail with @ChaelKruip. I do think that you can merge the technologies with the nodes and vice versa, skipping the `technologies` column.

grdw:
@antw  For now I just migrated the `technologies` column to the `technology_profile` column and leaving the `technologies` column empty (with an empty array)(this can be automatically filled with the data from the `technology_profile`, once we've decided whether or not to keep or remove this column).

antw:
Thanks @grdw! Looks good, so I went ahead and merged this.

Issue #85, opened 2015-04-28
Add Authorization to the Application


dennisschoenmakers:
Since some *Testing Grounds* can be very sensitive data, it should be prevented that other users could view any of the information associated with these Testing Grounds.

Also: there should be administrative users that can view everything. It should only be @ChaelKruip  to start with.

#### Question:
* Should we use Pundit or CanCan(Can)?

ChaelKruip:
:+1:

grdw:
@dennisschoenmakers do we really need to include a full gem in order to do something basic like  `if @user.admin?` and `if current_user == @testing_ground.user`?

antw:
> do we really need to include a full gem in order to do something basic like if @user.admin? and if current_user == @testing_ground.user?

For the moment, probably not. Maybe we'll want finer-grained control later, and a proper authorisation will be worthwhile then.

dennisschoenmakers:
> @dennisschoenmakers do we really need to include a full gem in order to do something basic like if @user.admin? and if current_user == @testing_ground.user?

Hmmm, I am afraid we will litter the app with these conditions, removing them later will be more painful, and I am getting convinced we will need it in the end anyways. So I tilt towards implementing a good solution straight away.

antw:
> and I am getting convinced we will need it in the end anyways

Fair enough. :+1:

antw:
While we're at it, I'm curious as to the justification for Monban over Devise. I know Devise isn't the new hotness, but it appears more actively maintained, gives us other useful features out-of-the-box (user activation, validation, forms), and will probably integrate easily with Pundit or CanCan. I'm open to persuasion, but at the moment I'm not really getting it.

dennisschoenmakers:
Monban... yes, @grdw: convince us! :smiley:

dennisschoenmakers:
![](http://upload.wikimedia.org/wikipedia/en/0/02/Statler_and_Waldorf.jpg)

grdw:
It's tiny, it's readable, it does the absolute basics of signing in a user (which is what we wanted ;) ). That's the only reason. I find devise big, invasive and horrible to maintain. Besides it's only cool to use devise when you want everything exactly as the devise gem intended to. If you want something custom (which, from experience, is usually the case in a future scenario) than you're screwed.

grdw:
@dennisschoenmakers and I were discussing lots of things considering authorization.

# The Basics
#### Private/public distinction between testing grounds. 

[...]

PR #89, opened 2015-05-04
Issues/#77 no auth notice

grdw:
Also contains a transition from Monban to Devise as discussed in #85


grdw:
Merged by hand - didn't think the changes are that big. You can still review it though.

Issue #99, opened 2015-05-11
Load curve info should be updated


ChaelKruip:
![ivy](https://cloud.githubusercontent.com/assets/1303760/7561813/4321a3fe-f7cf-11e4-9e17-71f100d6377b.png)

IMO the curve size is not interesting and could be removed.
I would like to see added:
* time-resolution (hours or 15 minutes or ...)
* category (in the tree)
* associated (allowed) technologies

Issue #98, opened 2015-05-11
Turn new testing ground form profiles into an editable table instead of a yaml input field


grdw:
As discussed

PR #101, opened 2015-05-12
Issues/#98 turn profiles into a table

grdw:
Contains commits for
#98, #99


app/assets/javascripts/profile_table.js
```diff
@@ -0,0 +1,77 @@
...
+
+  function parseTableToJSON(){
+    $("#testing_ground_technology_profile").text(tableToProfile());
+  };
+
+  function tableToProfile(){
+    var technologyProfile = _.map(tableRows(), function(tableRow){
+      return rowToTechnologyObject(tableRow);
+    });
+    return JSON.stringify(_.groupBy(technologyProfile, 'node'));
```
antw: Is Underscore really needed? It's a big dependency for only three uses. `_.groupBy` is pretty simple, and `Array.prototype.map` [is well-supported now](https://kangax.github.io/compat-table/es5/#Array.prototype.map), so the two instances of `_.map` can probably be replaced.

```javascript
function groupBy(elements, key) {
    var grouped = {},
        length  = elements.length,
        i, element;

    for (i = 0; i < length; i++) {
        element = elements[i];

        if (! grouped.hasOwnProperty(element[key])) {
            grouped[element[key]] = [];
        }

        grouped[element[key]].push(element);
    }

    return grouped;
}

groupBy([
    { category: 'main', name: 'First' },
    { category: 'rest', name: 'Second' },
    { category: 'main', name: 'Third' },
    { category: 'rest', name: 'Fourth' }
], 'category');
```

antw: I should add... I have no problem pulling in Underscore provided we're actually going to use it more. :smiley:

grdw: Totally agree with the underscore dependency. The less dependencies the better.

The `map` was because `$.map` completly didn't do what I expected, maybe 'plain' map will be a better option. I'll look into this. For now I could just copy the `groupBy` and use the `plain` map, and remove underscore.

grdw: Removed underscore.js with success :+1:

antw:
Just tried out the front-end... that is *nice!*

I think there are some interesting possibilities for this, such as making the "type" and "profile" selections into drop-downs, and I worry a bit that it's not obvious that the table can be clicked on to make changes. But, I think it's already *huge* usability improvement; thanks @grdw!

ChaelKruip:
> making the "type" and "profile" selections into drop-downs

I really like that idea!

> I worry a bit that it's not obvious that the table can be clicked on to make changes

@grdw and myself have discussed this as well. After next weeks workshop we can start to improve on it but I'm happy we have something that works :smile:

PR #115, opened 2015-05-13
adding house holds to the testing ground technologies list

### LINKED ISSUE #109: Automatically add houses + their respective base loads to the load profiles
As discussed with Chael

We're not adding houses for scaled-to-x-'inhabitants'-scenarios due to inhabitants per house.

grdw:
Fixes #109

### DIFF
--- a/app/models/installed_technology.rb
+++ b/app/models/installed_technology.rb
@@ -9,9 +9,10 @@ class InstalledTechnology
   attribute :capacity, Float
   attribute :demand,   Float
   attribute :storage,  Float
-  attribute :units,    Float,  default: 1.0
+  attribute :units,    Float, default: 1.0
 
-  EDITABLES = %i(name type profile capacity storage units)
+
+  EDITABLES = %i(name type profile capacity storage demand units)
 
   # Public: Returns a template for a technology. For evaluation purposes
   def self.template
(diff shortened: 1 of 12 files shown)


grdw:
@antw  could you take a look, if you have time. Thanks in advance!

app/models/installed_technology.rb
```diff
@@ -27,7 +28,7 @@ def self.template
   #
   # Returns true or false.
   def exists?
-    type.blank? || type == 'generic'.freeze || Technology.exists?(key: type)
+    type.blank? || type =~ /generic|base_load/ || Technology.exists?(key: type)
```
antw: This regex – and the same one in `Technology` – seems too permissive; it would match *any* technology with "generic" or "base_load" anywhere the key. I suggest changing it to `\A(?:generic|base_load)\Z`. Maybe make it a constant on `Technology` so that it doesn't need to be repeated?

grdw: I'd rather use `/^(generic|base_load)$/` which is the same as ? `\A(?:generic|base_load)\Z`. I use the 'general' word boundaries so that keys as `test_generic` and `generic_test` will fail.

grdw:
I changed the Regex and turned it into a constant, as suggested :)

antw:
> I'd rather use `/^(generic|base_load)$/` which is the same as ?

`(?:)` is a non-capturing group. However, I just tested them both and the performance difference is so negligible that I prefer the clarity of your capturing version.

> I use the 'general' regex boundaries so that keys as test_generic and generic_test will fail.

`\A` (start of string) and `\z` (end of string) [should be preferred](http://stackoverflow.com/a/577675) unless you *really* mean `^` (start of *line*) and `$` (end of *line*). That StackOverflow reply points out that I should have said `\z` and not `\Z`. TIL.

* `^` / `$`: http://rubular.com/r/fELyWJDjAj
* `\A` / `\z`: http://rubular.com/r/K2ULNjQJO0

Issue #130, opened 2015-05-15
Concurrency code needs some refactoring


ChaelKruip:
As discussed with @grdw, the current implementation of the concurrency calculation contains a lot of unnecessary duplications of objects and profiles. This will hurt performance badly for larger scenario's.

PR #136, opened 2015-05-19
Fixing the performance of the initial profile scheme with help of divmod

grdw:


grdw:
Fixes #130

antw:
Looks great, thanks!

PR #147, opened 2015-05-27
Issues/profile table select boxes

grdw:
Replaces the type column with a select box.

### DIFF
--- a/app/assets/javascripts/profile_table.js
+++ b/app/assets/javascripts/profile_table.js
@@ -27,11 +27,23 @@ var ProfileTable = (function(){
     var technologyObject = {};
     $.each(tableRow, function(i, techAttribute){
       var header = tableHeader(i);
+
+      if(header == "name"){
+        technologyObject["type"] = getTypeForName(techAttribute);
+      };
+
       if(!(/demand|capacity/.test(header) && techAttribute == "")){
         technologyObject[header] = techAttribute;
-      }
+      };
     });
     return technologyObject;
+  };
+
+  function getTypeForName(name){
+    var selectedOption = $("select.name").find("option").filter(function(){
+      return $(this).text() == name;
+    });
+    return selectedOption.val();
   };
 
   function tableHeader(index){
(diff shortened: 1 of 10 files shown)


grdw:
@antw if you have time to merge this. If you don't please tell, than I'll merge it myself.

antw:
Feels like a big improvement, but the behaviour of the drop-downs is a bit awkward. The first time I click on the table cell, the select appears (and is focused) but the options don't; I have to click a second time. 

It would also be nice for the profile select to update as soon as you change the technology type – currently I find that it only changes once I unfocus the technology type select.

Maybe these are limitations of the table editor library? Would be a pity because I think it will quickly get annoying to have to double-click everything.

grdw:
Thanks for your comments. Agree with the double-click, considering that I already made a lot of changes to the existing library, a little more changes might not hurt too much.

grdw:
> Feels like a big improvement, but the behaviour of the drop-downs is a bit awkward. The first time I click on the table cell, the select appears (and is focused) but the options don't; I have to click a second time. 

This is currently not possible, only with some dispatch-MouseEvent-Javascript voodoo which only works in Chrome. So it's kind of a bummer (it seemed like such an easy thing to do) :disappointed: 

>It would also be nice for the profile select to update as soon as you change the technology type – currently I find that it only changes once I unfocus the technology type select.

Done :smile:

antw:
> This is currently not possible, only with some dispatch-MouseEvent-Javascript voodoo which only works in Chrome. So it's kind of a bummer (it seemed like such an easy thing to do in the first place)

That's a shame; I think having to double-click the select box is poor user experience. However, it's still better than having to remember the names of technologies and profiles, so I think it's worth keeping.

I wonder if we could (eventually) replace the table-editor with a simpler solution where the table cells contain the form fields already, instead of being inserted dynamically by JS?

grdw:
> I wonder if we could (eventually) replace the table-editor with a simpler solution where the table cells contain the form fields already, instead of being inserted dynamically by JS?

That would be easier. Also saves a lot of JS; I'll discuss this with @ChaelKruip

PR #151, opened 2015-05-29
Auth/private public testing grounds

grdw:


db/migrate/20150529122843_add_permissions_to_topologies_load_profiles.rb
```diff
@@ -0,0 +1,8 @@
+class AddPermissionsToTopologiesLoadProfiles < ActiveRecord::Migration
+  def change
+    add_column :topologies, :permissions, :string, after: :graph, default: 'public'
```
antw: Do we expect permissions other than "public" and "private"? If not, I think this could just be a boolean column.

ChaelKruip: I think a boolean will do :+1:

PR #166, opened 2015-06-11
Issues/concurrency per tech

grdw:


antw:
The only issue I notice is that, when importing a scenario from ETEngine, the technologies table is (rightly) displayed with "max" concurrency. However, the radio buttons all default to "min". I think it would be better for the radio buttons to default to "max", since that reflects the state of the table when the page loads.

Otherwise, I think this looks good and can be merged :+1:

grdw:
@antw Fair point, I flipped the switched to 'max' :+1:

PR #167, opened 2015-06-11
Cleaning up the brush 'saving' a little bit

grdw:


grdw:
@antw could you look into this merge request :smile:  thanks in advance.

PR #184, opened 2015-06-23
Adding a custom error message when using a non existent profile

### LINKED ISSUE #149: User should get a clear error message when using a non-existent profile
Now ETMoses just tells you that:
ivy

Which should read "LES or Local Energy Solution" BTW

grdw:
Fixes #149


antw:
:+1:

PR #185, opened 2015-06-23
Setting defaults when changing the technology select box

### LINKED ISSUE #171: When changing a technology in 'profile table' set defaults
For instance:
- capacity
- storage volume
- etc.

grdw:
Fixes #171


antw:
@grdw Would it be possible to use the values from ETEngine as the defaults, instead of defining them in the database? It would be more resistant to changes in our data – it's annoying to have to manually update applications when a value in our dataset changes – and would better account for the possibility of different countries having different default values. Maybe it would be best to ask @ChaelKruip if you think it will take a lot of work.

I think the `(-)` and `(+)` buttons are missing an `event.preventDefault()` somewhere, because clicking them causes the page to jump to the top.

Otherwise, this looks good to me. I'm glad we've been able to switch to a simpler home-made table, as that solves those annoying usability issues with <select> fields. :+1:

ChaelKruip:
> @grdw Would it be possible to use the values from ETEngine as the defaults

Good point @antw!

grdw:
> @grdw Would it be possible to use the values from ETEngine as the defaults

@antw yes it would be possible :+1: 
I'll look into it.

grdw:
Now it takes the defaults based of the values from et-engine. 

PS: I didn't create another migration to remove the `defaults` column from the `technologies` table, but just removed it from the table and the migration-date from the `schema_migrations` table.

grdw:
> I think the (-) and (+) buttons are missing an event.preventDefault() somewhere, because clicking them causes the page to jump to the top.

Shortcut for this is using `#!` instead of `#` in the `link`

See http://stackoverflow.com/questions/1601933/how-do-i-stop-a-web-page-from-scrolling-to-the-top-when-a-link-is-clicked-that-t

antw:
Maybe I'm being thick, but why not `event.preventDefault()`? `#!` [is not a perfect solution](https://stackoverflow.com/questions/1601933/how-do-i-stop-a-web-page-from-scrolling-to-the-top-when-a-link-is-clicked-that-t#comment45685069_11246131).

```diff
diff --git a/app/assets/javascripts/profile_table.js b/app/assets/javascripts/profile_table.js
index 8760212..dd2a1a7 100644
--- a/app/assets/javascripts/profile_table.js
+++ b/app/assets/javascripts/profile_table.js
@@ -112,18 +112,22 @@ var ProfileTable = (function(){
   };
 
   function addClickListenersToAddRow(){
-    $("#profiles-table table tr th a.add-row").on("click", function(){
+    $("#profiles-table table tr th a.add-row").on("click", function(event){
       var row = $(this).parents("tr");
       var clonedRow = row.clone(true, true);
       clonedRow.find("td.editable:not(.select)").text("");
       clonedRow.insertAfter(row);
+
+      event.preventDefault();
     });
   };
 
   function addClickListenersToDeleteRow(){
-    $("#profiles-table table tr th a.remove-row").on("click", function(){
+    $("#profiles-table table tr th a.remove-row").on("click", function(event){
       $(this).parents("tr").remove();
       parseTableToJSON();
+
+      event.preventDefault();
     });
   };
```

Ideally these would not be `<a>` elements at all, since they don't function if JS is disabled (in fairness, nor does any other part of the application :laughing:). Same goes for the "Toggle all" and "Apply" (which could be changed to `<button>`s if `preventDefault` is used).

grdw:
Point made.

PR #186, opened 2015-06-23
Issues/180 sorting of profile classes

grdw:


app/services/load_profiles/hierarchy.rb
```diff
@@ -0,0 +1,44 @@
...
+    #
+    # Query to sort Load Profile Categories hierarchly
+    #
+
+    def initialize
+    end
+
+    def tree_sort
+      with_paths.sort do |category_a, category_b|
+        category_a[:path] <=> category_b[:path]
```
antw: Maybe this could be...

```ruby
def tree_sort
  with_paths.sort do |category_a, category_b|
    sort_criteria(category_a) <=> sort_criteria(category_b)
  end
end

private

def sort_criteria(category)
  [category[:path], category[:load_profile_category].name.downcase]
end

```

... which would sort subcategories at the same level by their name?

grdw: Fixed it a little differently due to the fact that the paths are all unique. See line 32.

antw:
@grdw I think I must be doing something wrong:

* Sorting seems to be not-quite-right to me...

  screen shot 2015-06-26 at 14 07 58

  I put the parent category of each sub-category in brackets. The sub-sub category seems to be placed correctly, but the others are not.

  screen shot 2015-06-26 at 14 09 19

* If a user is editing a load profile, the `<select>` is just selecting the first category in the list, instead of the one the profile uses.

--

I have pushed a branch, `issues/180-tree-sorting-of-profile-categories`, with a different take on this. It'll sort the categories into a tree of hashes, and then recurse through that to create the `<option>`s. [The commit is here](https://github.com/quintel/etmoses/commit/1e749b2a72c19b9d626763766c09b8b7986b5d9c).

If you can think of a better way – I think mine is a bit ugly – please feel free to go in another direction. :smiley: 

screen shot 2015-06-26 at 15 12 24

grdw:
Oops, I only sort by the first letter ... 

I changed

`category.name.first` to `category.name.downcase` on line 32.

Maybe that sorts it better.

grdw:
I think I fixed it. But it's friday so my judgement is way below the standard. If you don't mind I'll look friday. I gave the 'Main category's' the `parent_id`, `0`

grdw:
@antw I tested it a bit more and I think this is the working version. `parent_id` is a mandatory field and if you want to specify a top-level category you have to change the `parent_id` to `0`.

antw:
Looks good now, thanks! :+1:

PR #192, opened 2015-07-01
Market models/import table

grdw:


app/models/profile.rb
```diff
@@ -0,0 +1,59 @@
+class Profile < ActiveRecord::Base
+  include Privacy
+
+  belongs_to :user
+
+  has_attached_file :curve, styles: {
```
antw: @ChaelKruip @grdw I'm guessing the new FinancialProfile model is intended to allow us to upload profiles to specify pricing in a market model?

Do we know yet what the contents of these profiles will look like? I ask because I'm pretty sure the demand/capacity scaling processors used on load curves won't apply to financial curves. I'm think we can move this to `LoadProfile` and add another `has_attached_file` to `FinancialProfile`? That's assuming Paperclip works with STI, but I don't see any reason it wouldn't...

ChaelKruip: > Do we know yet what the contents of these profiles will look like?

They will be profiles containing numbers in EUR which don't need to be scaled indeed :smile:

app/models/market_model.rb
```diff
@@ -0,0 +1,17 @@
+class MarketModel < ActiveRecord::Base
+  include Privacy
+
+  FOUNDATIONS = ["kWh", "kW_connection", "kW_max", "kW_contracted", "kW_flex"]
```
antw: I think we're going to start out with just one of these, and then add the others as the sprint progresses. I think we should be starting with `kW_max` or `kWh` but I haven't decided yet (it'll be whichever is easiest :laughing:). It's okay for us to keep the full list so long as the client understands that only one of them will work for the time being; otherwise we should probably narrow it down to "kW_max".

app/models/profile.rb
antw: > They will be profiles containing numbers in EUR 

Maybe we should call it `PriceCurve` (for consistency with [merit](https://github.com/quintel/merit/blob/00abba2c5c0ce5c627cfd00ab7451d2f8822507d/lib/merit/order.rb#L87-L94)), or `PriceProfile`? It's not really a *profile*...  a profile describes *shape* while the price curve has real values. (I'm just arguing semantics here, this isn't important :smiley:)

```
Curve
 ├ PriceCurve
 └ LoadProfile
```

app/models/market_model.rb
ChaelKruip: I actually think that 'kWh' and 'kW_connection' are going to be the most important (kW_max and kW_contracted are currently only applicable to larger users, say industry). Luckily, I also think these are pretty simple:
* kWh: (load * time) consumed per customer over the whole year.
* kW_connection: a fixed price per year per customer. This is mostly 3x25 Amps. Sometimes 3x80 Amps. We could either make it the responsibility of the user to assign to each end-point of the topology a capacity of either of these two types. The disadvantage of that is that there could be more or fewer households connected to that end-point than the capacity was intended for and the user would have to adjust the topology for every LES.

antw: > I actually think that 'kWh' and 'kW_connection' are going to be the most important 

Fine with me. :+1: kWh is probably the simplest of the two, so I think I'll start there.

antw:
Finally got around to trying out the market model form, and I like what you've done: it'll be familiar to users who have used the testing ground form, and the drop-downs for tariffs is smart.

Do we need to connect a market models to a topology somehow to get the stakeholder to/from options to show something?

screen shot 2015-07-01 at 16 40 51

grdw:
I added stakeholders as a table with a `name` attribute only. I'm still doubting if they have to be in the database or if a simple array with names will do.

antw:
> I added stakeholders as a table in the database with a name attribute only. I'm still doubting if they have to be in the database or if a simple array with names will do.

Sorry, I was being stupid; I don't know why I thought it would need to be connected to the topology to show the stakeholders. :smiley:

At this point, stakeholders don't seem to have any special behaviour, so I expect an array would work fine.

grdw:
> At this point, stakeholders don't seem to have any special behaviour, so I expect an array would work fine.

I changed it into a simple array and removed the table :)

antw:
I'll hold off merging until a decision is made on #197. Looks good though. :+1:

Issue #178, opened 2015-06-22
Treat EDSN profiles differently from individual profiles when assigning them to households


ChaelKruip:
EDSN profiles represent groups of households larger than typically 100. It would be nice if they are selected only for connections with > 100 households.

We have to think what would happen if the concurrency is minimised for such a group...

grdw:
I think we should handle it as a 'separate technology'. Household larger than 100 and Household's smaller than 100. It would than automatically select the correct 'technology' according to the total amount of `units`.

ChaelKruip:
> It would than automatically select the correct 'technology' according to the total amount of units.

As long as that is the case you can make it two different technologies indeed. Not highest priority now.

grdw:
I'm well on my way with this issue. Though we still need to think about the concurrency.

ChaelKruip:
> Though we still need to think about the concurrency.

If after minimising concurrency the groups become smaller then the 'threshold' for aggregation, the individual profiles should be used IMO. 

Also, I think that the threshold should be lower than 100. Perhaps we can even make this user-defined?

ChaelKruip:
> Though we still need to think about the concurrency.

If after minimising concurrency the groups become smaller then the 'threshold' for aggregation, the individual profiles should be used IMO. 

Also, I think that the threshold should be lower than 100. Perhaps we can even make this user-defined?

grdw:
In merge #206

PR #206, opened 2015-07-08
Issues/#189 capping household profile selection

### LINKED ISSUE #189: Users should be able to specify how much of solar PV production can be capped
Perhaps we can add this to the table?

- Insert slider from 0-100% - for capping.

### LINKED ISSUE #201: Toggle for concurrency options should toggle between 'all min' and 'all max'
Now it really toggles everything the user specified before... This can be quite annoying because the user has to manually reset the buttons.

Also (as suggested by @dennisschoenmakers): switches might be more intuitive than the current 'horizontal radio buttons'.
energietransitiemodel_-_je_eigen_onafhankelijke__volledige_en_op_feiten_gebaseerde_energiescenario_

grdw:
- Fixes #201 
- Fixes #189


antw:
Looks good. I don't think it closes #189 – perhaps you meant #178?

#189 means to add a % slider (0-100%) to control how much solar PV can be capped when there isn't sufficient demand for all the energy. Doesn't matter to me that it's not included, though, as the network calculation doesn't yet model this behaviour.

grdw:
I mean #178 yes. My bad.

PR #215, opened 2015-07-09
Extracting load profiles to load curves

grdw:
- Please merge #206 first
- After merging I need to do some manual changes to some profiles (so let me deploy - after it's merged) :)


.gitignore
```diff
@@ -47,3 +47,4 @@ bower.json
 # Temporary Paperclip test files.
 /public/test_files
 .zedstate
+.DS_Store
```
antw: You might want to add .DS_Store to [your global ~/.gitignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) so that you don't need to do it in every project.

db/migrate/20150703084923_move_existing_load_profiles_to_load_curves.rb
```diff
@@ -0,0 +1,14 @@
+class MoveExistingLoadProfilesToLoadCurves < ActiveRecord::Migration
+  def change
+    LoadProfile.all.each do |load_profile|
```
antw: Migration fails on af0ca565 as the `LoadProfile` class is now `ProfileCurve`. I think you'll have to drop down to raw SQL to run this migration now. :persevere: 

```
StandardError: An error has occurred, all later migrations canceled:

NameError: uninitialized constant MoveExistingLoadProfilesToLoadCurves::LoadCurve
moses/db/migrate/20150703084923_move_existing_load_profiles_to_load_curves.rb:5:in `block in change'
moses/db/migrate/20150703084923_move_existing_load_profiles_to_load_curves.rb:3:in `change'
Tasks: TOP => db:migrate
```

grdw:
I'm going to take a closer look to the database migrations.

grdw:
I changed the migrations. So in order to run those you need to start out with the same database as is currently on the staging environment. After that you run all the migrations and they should succeed. If not, please contact me.

antw:
Uploading a new load profile – without manually specifying a key on the form – yields this error:

```
NameError - undefined local variable or method `curve_file_name' for #<LoadProfile:0x007ff6a9f343d0>:
  activemodel (4.1.8) lib/active_model/attribute_methods.rb:435:in `method_missing'
  activerecord (4.1.8) lib/active_record/attribute_methods.rb:213:in `method_missing'
  app/models/profile.rb:11:in `block in <class:Profile>'
```

* [LoadProfile:15](https://github.com/quintel/etmoses/blob/b4ea80b2d07f01196450463e0f206e3bab502ec4/app/models/profile.rb#L15)

Uploading a financial profile gives this error:

```
Paperclip::Error - FinancialProfile model missing required attr_accessor for 'curve_file_name':
  paperclip (4.2.1) lib/paperclip/attachment.rb:406:in `block in ensure_required_accessors!'
  paperclip (4.2.1) lib/paperclip/attachment.rb:404:in `ensure_required_accessors!'
  paperclip (4.2.1) lib/paperclip/attachment.rb:98:in `assign'
  paperclip (4.2.1) lib/paperclip/has_attached_file.rb:66:in `block in define_setter'
```

I'm pretty sure I'm fully-migrated. I started with an export of the database on staging, and ran the migrations.

--

I'm thinking that `FinancialProfile` shouldn't be a subclass of `Profile`, but a separate class entirely (like `ProfileCurve`). If I'm understanding your implementation correctly, `Profile` acts like a group of curves rather than a single curve. If the financial and load curves share common methods, they can always be placed into a module.

At the risk of being a real pain in your side (sorry!): there's some confusion about naming. Curves which describe *load* are normally called "profiles" because they don't have real load values in them, but instead they show the "shape" of that load over time (their values are more like "10%", "5%", ...). The real load curve is a function of the profile and the technology `capacity`. The financial curves *do* contain real values (prices in euros), so these would be called curves rather than profiles.

FinancialProfile → PriceCurve
Profile → LoadProfile
ProfileCurve → LoadProfileComponent

I know this seems like pedantry, but [other libraries](https://github.com/quintel/merit/blob/00abba2c5c0ce5c627cfd00ab7451d2f8822507d/lib/merit/participants/producer.rb#L9) use this naming convention and it risks confusion if we aren't consistent.

ChaelKruip:
> I know this seems like pedantry, but other libraries use this naming convention and it risks confusion if we aren't consistent.

I agree that this is important. Thanks for being precise about this @antw. Sorry @grdw for not being so precise when explaining this to you earlier.

grdw:
I'll handle the renaming. And this feature needs more specs and testing ...

grdw:
> FinancialProfile → PriceCurve
> Profile → LoadProfile
> ProfileCurve → LoadProfileComponent

@antw to be clear.

Your suggestion is:

- Rename the `profiles` table to `load_profiles`
- Rename the `profile_curves` table to `load_profile_components`
- Make another table for `PriceCurve` (`price_curves`)

antw:
> Rename the profiles table to load_profiles

Yes, I *think* so.The profiles are specific to load rather than being generic curves which describe shape, so I think it makes sense to call them *Load*Profiles.

> Rename the profile_curves table to load_profile_components

My reasoning here is that the "components" describe different parts of the total load profile; that is, for base loads, there's a flexible part and an inflexible part which together make up the complete profile.

> Make another table for PriceCurve (price_curves)

Maybe PriceCurve and LoadProfile could have a common ancestor ("Curve") but, personally, I'd lean towards them being separate. I don't have a strong opinion so I don't mind either way.

grdw:
@antw I renamed the tables accordingly. And made a tiny module called `CurveComponent` which contains the `as_json` and `merit_curve` methods that previously only existed in `LoadProfileComponent`.

antw:
Thanks @grdw! I think that's better.

Unfortunately I think the class name change has broken a migration again as `load_profiles` was already renamed to `profiles` (I'm migrating a copy of the current "staging" DB):

```
== 20150703084923 MoveExistingLoadProfilesToLoadCurves: migrating =============
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

Mysql2::Error: Table 'etloader_development.load_profiles' doesn't exist: SELECT `load_profiles`.* FROM `load_profiles`
moses/db/migrate/20150703084923_move_existing_load_profiles_to_load_curves.rb:3:in `change'
ActiveRecord::StatementInvalid: Mysql2::Error: Table 'etloader_development.load_profiles' doesn't exist: SELECT `load_profiles`.* FROM `load_profiles`
moses/db/migrate/20150703084923_move_existing_load_profiles_to_load_curves.rb:3:in `change'
```

I think `has_attached_file` can be removed from LoadProfile since it was moved to LoadProfileComponent?

Thanks for what you've done in this PR; the end result is much clearer than what we had before. :+1:

grdw:
> I think has_attached_file can be removed from LoadProfile since it was moved to LoadProfileComponent?

Yes

I fixed the `LoadProfile --> LoadProfileComponent` migration rather hacky, though it worked on my local machine after `scp`'ing all the csv-files from the server.

antw:
Perfect, thanks! :+1:

PR #225, opened 2015-07-15
Issues/#212 flexibility strategies availability

grdw:


antw:
Thanks @grdw; I'll review this properly tomorrow.

app/models/network/battery.rb
```diff
@@ -7,5 +7,9 @@ module Network
   # inherit from the Storage class.
   class Battery < Storage
     extend ProfileScaled
+
+    def self.disabled?(options)
+      options[:battery_storage] == 'false'
```
antw: I'm not keen on `false` being a string. I'm guessing this is coming from the `params` hash, but the Rails-side of the application should be casting these values to `true` or `false` before passing them into the Network module.

The Network and Market namespaces will likely be extracted out of Moses and into a separate Gem, so it's a good practice to already treat it like an external library.

app/models/load_profile.rb
```diff
@@ -9,6 +9,11 @@ class LoadProfile < ActiveRecord::Base
 
   validates :key, presence: true, uniqueness: true
 
+  has_attached_file :curve, styles: {
```
antw: Did you mean to add this back in? I understood this had moved into LoadProfileComponent?

app/models/network/conserving_producer.rb
```diff
@@ -5,18 +5,27 @@ class ConservingProducer < Technology
...
+      production > capping ? capping : production
     end
 
     def store(_frame, _amount)
     end
+
+    private
+
+      def capping_percentage
+        options['capping_solar_pv_slider_value'].to_f
```
antw: Same here as above with `"false"`. The Rails-side should be casting these values before creating the Network (in fact, probably before saving them to the database).

app/models/network/technology.rb
```diff
@@ -29,14 +29,15 @@ def self.behaviors
...
     def self.build(installed, profile, options)
-      new(installed, profile)
+      new(installed, profile, options)
     end
 
-    attr_reader :installed, :profile
+    attr_reader :installed, :profile, :options
 
-    def initialize(installed, profile)
+    def initialize(installed, profile, options)
```
antw: Instead of storing the entire `options` hash on every Technology, it's likely better to extract the values you care about and ignore the rest.

At the *very* least, `Hash#fetch` should be used so that an exception gets raised if an option key is missing:

```ruby
def initialize(installed, profile, options)
  super
  @capping_fraction = options.fetch(:capping_solar_pv_slider_value)
end
```

However, since we're targeting Ruby 2.1 and newer, I think this would be a good case for using keyword arguments. Classes which need to accept an option can do it with a keyword argument instead:

```ruby
class ConservingProducer
  def initialize(installed, profile, capping_fraction:, **)
    super
    @capping_fraction = capping_fraction
  end

  # ...
end
```

In the example, I called the option `capping_fraction` rather than `capping_solar_pv_slider_value` because the latter is a front-end term. Be as generic as possible when naming thing in the back-end code, because it may be reused for other purposes. You might want to have a wind turbine whose production is capped, and "solar_pv_slider_value" doesn't make much sense in that context.

app/models/network/conserving_producer.rb
```diff
@@ -5,18 +5,27 @@ class ConservingProducer < Technology
...
+      production = super
+      capping = (capacity * capping_percentage).abs
+
+      production > capping ? capping : production
     end
 
     def store(_frame, _amount)
     end
+
+    private
```
antw: Last comment for tonight (I swear!).

The normal QI Ruby style is [not to indent](https://github.com/bbatsov/ruby-style-guide#consistent-classes) after calling `private`, `protected`, or `module_function`; just maintain the same level of indentation as before. There's a RuboCop [style config here](https://gist.github.com/antw/c5297c5e7d9138dc83da) – which is mostly consistent with the [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide) – but you probably don't want to run that now as there'll be too many messages (mostly fixed on that "markets" branch).

I think RuboCop will *always* complain about "inconsistent indentation" if you indent after `private`.

app/models/load_profile.rb
grdw: I'll remove this one.

app/models/network/conserving_producer.rb
grdw: The indentation is something inherited from my previous company. And I'll change it.

**Another question about `private` and style:**
The  `#######`,  surrounding `private` - which I see a lot - isn't mandatory right? Because I don't really understand the purpose of it. I can grasp the idea of "whenever a `class` is growing to large, you can more easily see where the `private` methods begin". Though, when your `class` is to large, it's doing too much.

antw: No, not mandatory at all. It's a personal habit, and probably one I should stop. :laughing:

grdw:
@antw could you take another look please?

PR #230, opened 2015-07-16
Issues/#226 include stakeholders into topology

### LINKED ISSUE #226: Including stakeholders into topology
- [x] Adding validation for stakeholders in Topology model
- [x] Showing stakeholders sub-menu section
- [x] When clicking a stakeholder - show which topology-nodes include this stakeholder.

grdw:
FIxes #226


antw:
Merged since @ChaelKruip wants to test this tomorrow, and it looks fine to me.

One thing I noticed is that each node may belong to a single stakeholder. I'd like the node to be able to belong to multiple stakeholders, i.e. change the attribute from a string into an array of strings.

This will be generally useful for the market models, but can be used to allow an endpoint to be both a customer and a member of a co-operation.

Issue #219, opened 2015-07-13
Which "Measures" for market models do we need


ChaelKruip:
This is an attempt to gather all info about the 'measures' we will need to realise in order to calculate the relevant business cases for our market models. We may decide not to include some of the measures described below, or perhaps we later think of others later so this list is not definite.

I think of a 'measure' as the quantification of some property in the LES that can serve as the basis for a transaction. So, the number of kWh's of electricity consumed over the year by the 'Customer' stakeholder would be a simple example.

When thinking about measures, I realised that most of them are related to either the 'Customers' or the 'Co-operation' stakeholders. Both these stakeholders are actually groups of end-points of the topology including all attached technologies and base_loads. So, most of the measures happen on the end-nodes. Unless specified differently, each measure applies to both 'Customers' and the 'Co-operation'. Within the group of connections that is the 'Customer' or the 'Co-operation', the loads can be summed.

Measures fall in roughly four categories: capacity, commodity, flexibility and outages. 

# Capacity-related measures

- [x] The number of base_load profiles (this is how we determine how many 'user connections' there are to the network. Note that an end-point of the network can have many 'user connections' (the 'units' of the base_load).
- [ ] The yearly maximum peak of the sum of all (for the selected stakeholder) load on the lowest level of the network (this is basically, the 'kW_gecontracteerd')
- [x] The monthly maximum peak load (the 'kW_max')
- [x] The instantaneous load (i.e. a vector containing 35,040 points)
- [ ] Depreciation cost of all transformers in LES (later also cables): i.e. installation costs divided by its economic lifetime

# Commodity-related measures

- [x] kWh consumed
- [x] kWh produced. NOTE: At the moment we assume that all production at the lowest level is volatile (either wind or solar) or from discharging batteries (P2P and EV) which we also count as renewable
 - [ ] kWh produced for those who produce under *x* (e.g. 10,000) kWh on average (per connection) OPTIONAL
- [x] kWh produced by Producer (mix of installed plants determined by Merit Order) which is equal to the net load on the highest node of the network (HV)

NOTE: The following condition is in effect for these measures: no flow of energy (kWh) is possible through a transformer that is congested or has been congested in the past **three hours**. This means that all connection under a congested transformer cannot consume or produce energy. This holds even if the sub-net is capable of transporting that energy (the reason is that the 50Hz signal disappears when transformers break down and this signal is needed by all electrical technologies considered here).

# Flexibility-related measures

[...]

PR #232, opened 2015-07-16
Adding measure to market models

grdw:


antw:
[I don't think this is needed anymore](https://github.com/quintel/etmoses/issues/219#issuecomment-121141196), as foundations and measures are going to be combined into one object.

Sorry to have wasted your time, @grdw. I'll close this, but please re-open it if I've misunderstood something...

PR #242, opened 2015-07-22
Fixing IE issues with CSV reading

grdw:


antw:
Can we have a test for what this fixes?

I'm curious what the CSV library is doing, because maybe we will also need to change Merit::Curve (or [Network::Curve](https://github.com/quintel/etmoses/blob/fdd01e5f83e8024cba46dde953e4cba1ef9e2324/app/models/network/curve.rb) which is on the `markets` branch). Some technologies use the original curve which was uploaded by the user, so if Merit::Curve isn't sufficient in the Paperclip processor, it likely won't be good enough elsewhere either.

The ETCurve class ([or module](https://github.com/bbatsov/ruby-style-guide#modules-vs-classes)) could probably be merged into the Paperclip processor; a two-line class only used in one place is a bit redundant.

grdw:
I'll add tests, though it's hard to test without a Windows machine.

There's this case where the csv file has a line-break (`\r`) at the end of the file which tends to break the overall processing of the csv file. If you remove the last blank line in the csv file, all is fine.

grdw:
I hate to admit it. But the fix with reading the `CSV` file first makes the overall test suite slower by 4-5 seconds.

antw:
Bear with me a bit, I think I can fix the line ending issue in Merit.

grdw:
:+1:

antw:
Hopefully [fixed in Merit](https://github.com/quintel/merit/commit/d631d6178a2d5824d718de713effffe312250f53). Performance of `Curve.load_file` took a 30% performance hit, sadly. We could detect `\r` endings when the file is uploaded and convert them to `\n`, but maybe it's best that the Merit gem be more robust with this stuff (and we [cache the files in production](https://github.com/quintel/etmoses/blob/master/app/models/curve_component.rb#L11), anyway). For me, the Moses tests run in much the same time as they did before (~9.1s instead of ~8.8s).

grdw:
> For me, the Moses tests run in much the same time as they did before (~9.1s instead of ~8.8s).

This is with your `Merit` fix?

antw:
> This is with your Merit fix?

Yes. I benchmarked `Curve.load_file` before and after, and after came to about 30% slower. But I guess it's not being called very often, so it doesn't have much effect on the Moses tests.

grdw:
@antw is this already merged into the merit gem?

antw:
It should be on master already: b7425eaf.

PR #254, opened 2015-08-05
Issues/load management

grdw:


antw:
Sadly I'm thinking we may have some misunderstandings (or miscommunication) here. I'm not clear why the deferring of loads was removed from DeferredConsumer and moved into a separate calculator step (`ShiftConsumption`/`shifting_consumption!`); could you explain your reasoning behind this?

grdw:
The thing is that some specs - after some thorough discussion with @jorisberkhout - weren't making a whole lot of sense. In order to make them succeed I moved the code around.

antw:
What wasn't making sense?

I'll be in the office next Tuesday/Wednesday, so perhaps the three of us can discuss this in more detail in person? I'm still curious to know what was wrong though. :smiley:

grdw:
This spec: https://github.com/quintel/etmoses/blob/issues/load-management/spec/controllers/testing_grounds_load_management_controller_spec.rb#L151:L161

But to explain it in theory.

For instance these are my flex and inflex parts:
The node-capacity is 1.0
````ruby
  flex = [0.2, 0.0, 0.0, 0.0]
inflex = [1.0, 1.0, 0.0, 1.0]

# Total load =
[1.2, 1.0, 0.0, 1.0]

````

The thing that should happen with the postponing of the base load - as discussed with @jorisberkhout - is that the flex part (the `0.2`) should move 2 spaces.

Which would result in:

````ruby
# Total load =
[1.0, 1.0, 0.2, 1.0]
````

This is something I didn't saw happening in the original code.

antw:
@jorisberkhout @grdw Thanks for the Skype call; it's good to know we're on the same page as to how this should work.

I know what the issue is with the original DeferredConsumer implementation. Actually, the original class *and* the tests written by @grdw are both correct – which probably sounds ridiculous since they seem to disagree with one another. :laughing:

Tomorrow I'll make a couple of changes to this pull request (on another branch) and will explain...

PR #257, opened 2015-08-06
Saving and postponing for flexible base load components

antw:
My take on #254, this PR takes the original `DeferrableConsumer` with @grdw's changes to profiles and components to support load management features on base loads.

The tests originally written by @grdw were correct, but made it appear that DeferrableConsumer did not work. This was because the consumer expects to be given a `capacity`, and the tests provided `capacity=nil`. This resulted in it defaulting to a capacity of 0.0, and thus no load was placed on the network until 12 frames after the exceedance (instead of at the first available opportunity).

The class was designed for a "discrete" technology – which would be assigned a `capacity` – whereas base loads are instead assigned a `demand`. I have changed the class so that the technology capacity will default to infinity if none is set. It will now work as expected when used with a base load.

FWIW, there are a couple of reasons why these features need to be implemeted by DeferrableConsumer, and not as a separate post-processing step ([ShiftProduction](https://github.com/quintel/etmoses/blob/633e6d73bd83cd719618c8f9ffa4b05d55b51c3a/app/models/calculation/shift_consumption.rb)):

* Postponed loads need to respect the technology capacity, if one is set. DeferrableConsumer uses [CapacityLimit](https://github.com/quintel/etmoses/blob/ae808caabd5c7b16ca8da65c16e8c00c955a7c5d/app/models/network/capacity_limit.rb) which guarantees this.

* [The order in which technologies are assigned load](https://github.com/quintel/etmoses/blob/ae808caabd5c7b16ca8da65c16e8c00c955a7c5d/app/models/calculation/context.rb#L61-L69) is important, and ShiftProduction would cause postponing to always happen after all other loads have been assigned. In fact, it should be given priority over power-to-power batteries (which uses `Battery`), power-to-heat (implemented as `OptionalBuffer`), and power-to-gas (`Siphon`).

I also added a [TechnologyComponentBehavior model](https://github.com/quintel/etmoses/commit/ae808caabd5c7b16ca8da65c16e8c00c955a7c5d) so that we can define the behavior of the flexible component in the database, instead of us having to hard-code it somewhere.

I think this is now ready to be merged.

--
cc @jorisberkhout


grdw:
>  capacity, and the tests provided capacity=nil

So my tests were - in a way - incorrect :wink: 

This merge request looks good though :+1:

PR #258, opened 2015-08-10
Issues/#256 split up week and full view

grdw:


antw:
Discussed with Chael, and I'll close this since the redesigned front-end will replace these changes.

PR #262, opened 2015-08-11
Replacing load profile keys with id's

### LINKED ISSUE #261: Use load profile id's instead of specified 'keys'
Due to the fact that users don't have to type the technology data for a testing ground themselves and instead just use a table with select boxes. We might aswell just use the load_profile id's instead of the keys. This saves space and allows easier renaming of the keys without destroying alot of testing grounds.

grdw:
- Saves space in the database
- Makes renaming of the keys easier without breaking the whole database

Fixes #261


antw:
Thanks! I also changed InstalledTechnology#profile to return an integer instead of a string.

PR #272, opened 2015-08-12
Issues/#263 comparing business cases

grdw:


antw:
I think it would be useful to give the user more feedback in the event that creating a business case fails in the BusinessCaseController#find_testing_ground hook. This can fail if no market model is assigned to the LES, but no feedback is given to the user: it simply redirects back to the LES page and the user won't know what went wrong.

Otherwise, it seems to work fine and can probably be merged.

@ChaelKruip I can merge this now if it is needed for testing.

ChaelKruip:
> I can merge this now if it is needed for testing.

That would be nice!

Issue #324, opened 2015-08-26
Cancelling a change in the concurrency settings messes up the technologies


jorisberkhout:
While editing a LES, I accidentally pressed the `Apply` button under the concurrency switches. I cancelled the operation through the `Cancel` button on the pop-up but this messed up the technologies. I noticed the following changes:

- the residential PV panel was changed from `solar_pv_amsterdam` to `solar_pv_zwolle`
- the yearly demand for the residential PV was set to `0` where it had no value before the operation
- the latter also happened to the yearly demand for the power-to-heat that I had in my LES

I was expecting no effect since I cancelled the operation.

Issue #289, opened 2015-08-24
When changing concurrency back to max for households, the individual profiles should be replace by aggregated ones


ChaelKruip:
A group of 33 households should get an aggregated profile:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/9438854/0f9ada06-4a62-11e5-8f62-c34228c649c5.png)

grdw:
That's one of the problems I found when trying to implement #273

Issue #273, opened 2015-08-13
Making a separate technology for EDSN base loads


grdw:

grdw:
I made a seperate technology for edsn-base loads and gave it the `optional` behavior. 

@ChaelKruip Does the code (based upon the units) need to auto-detect which base load technology type there's needed (aggregated or anonimous) ?

ChaelKruip:
> @ChaelKruip Does the code (based upon the units) need to auto-detect which base load technology type there's needed (aggregated or anonimous) ?

Yes.

ChaelKruip:
Can this be closed @grdw?

grdw:
Not yet. There's still some front-end things that need to be discussed.

Issue #268, opened 2015-08-11
When editing a LES, the assignment of individual profiles for households is overwritten


ChaelKruip:
If the group is larger than 30, ETMoses does not seem to respect the choice of the user to assign an individual profile anyway.

PR #331, opened 2015-08-28
Edsn household changes

grdw:
Fixes https://github.com/quintel/etmoses/issues/324
Fixes https://github.com/quintel/etmoses/issues/268
Fixes https://github.com/quintel/etmoses/issues/289
Fixes https://github.com/quintel/etmoses/issues/260
Fixes https://github.com/quintel/etmoses/issues/273

### DIFF
--- a/app/models/installed_technology.rb
+++ b/app/models/installed_technology.rb
@@ -73,7 +73,8 @@ class InstalledTechnology
   def behavior_with_curve(curve_type = nil)
     behavior = self.behavior.presence || technology.behavior
 
-    return behavior if curve_type.nil? || curve_type == 'default'.freeze
+    return "default" if is_edsn_profile?
+    return behavior  if curve_type.nil? || curve_type == 'default'.freeze
 
     component_behavior = technology.component_behaviors.for_type(curve_type)
     component_behavior.try(:behavior) || behavior
@@ -118,6 +119,10 @@ class InstalledTechnology
     end
 
     multiplier * units
+  end
+
+  def is_edsn_profile?
+    profile =~ /edsn/
   end
 
   def ratio(component)
--- a/app/models/installed_technology.rb
+++ b/app/models/installed_technology.rb
@@ -73,8 +73,7 @@ class InstalledTechnology
   def behavior_with_curve(curve_type = nil)
     behavior = self.behavior.presence || technology.behavior
 
-    return "default" if is_edsn_profile?
-    return behavior  if curve_type.nil? || curve_type == 'default'.freeze
+    return behavior if curve_type.nil? || curve_type == 'default'.freeze
 
     component_behavior = technology.component_behaviors.for_type(curve_type)
     component_behavior.try(:behavior) || behavior
@@ -119,10 +118,6 @@ class InstalledTechnology
     end
 
     multiplier * units
-  end
-
-  def is_edsn_profile?
-    profile =~ /edsn/
   end
 
   def ratio(component)
--- a/app/assets/javascripts/concurrency.js.coffee
+++ b/app/assets/javascripts/concurrency.js.coffee
@@ -12,6 +12,7 @@ topology_id = ->
 
 technology_distribution = ->
   tech_distribution = JSON.parse($("textarea#technology_distribution").text())
+
   for tech in tech_distribution
     concurrency = $(".check_box input[name='" + tech.type + "']")
     tech.concurrency = (if concurrency.prop('checked') then 'max' else 'min')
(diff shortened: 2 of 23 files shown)


antw:
I'm not seeing any profiles available when importing a large (EDSN) household. I solved this by manually adding a `base_load_edsn` technology; can we add a migration to do this, or was this not the right thing for me to do?

However, toggling the concurrency on the `base_load_edsn` doesn't have any effect (`base_load` techs in a smaller testing ground does still work). For reference, the larger scenario (with EDSN profiles) is "425372", and the smaller one is "425532".

grdw:
> However, toggling the concurrency on the base_load_edsn doesn't have any effect (base_load techs in a smaller testing ground does still work).

@ChaelKruip and I discussed this and considered that it made little sense to change the concurrency for EDSN profiles.

grdw:
> I'm not seeing any profiles available when importing a large (EDSN) household. I solved this by manually adding a base_load_edsn technology; can we add a migration to do this, or was this not the right thing for me to do?

Yes. I'll add the migration.

antw:
> @ChaelKruip and I discussed this and considered that it made little sense to change the concurrency for EDSN profiles.

Okay! I assumed they'd be converted from `base_load_edsn` to `base_load` (and change the profiles accordingly), but not supporting concurrency on this tech seems fine also. Is it possible to hide the concurrency toggle for the EDSN tech, or will that be a bit hacky?

grdw:
> Is it possible to hide the concurrency toggle for the EDSN tech, or will that be a bit hacky?

Yes. I added a `visible` property to the technologies table which takes care of that.

app/models/installed_technology.rb
```diff
@@ -71,7 +71,8 @@ def technology
   def behavior_with_curve(curve_type = nil)
     behavior = self.behavior.presence || technology.behavior
 
-    return behavior if curve_type.nil? || curve_type == 'default'.freeze
+    return "default" if is_edsn_profile?
```
antw: I'm not sure I understand why an EDSN profile means `default` behavior (btw, the correct name for the "featureless" behavior is `generic`). Shouldn't it be `default` for the inflexible component and `optional` (saving) for the flexible? If so, this would be accomplished by creating two new TechnologyComponentBehavior records:

```ruby
tech = Technology.find_by_key('base_load_edsn')

tech.component_behaviors.create!(curve_type: 'flex',   behavior: 'optional')
tech.component_behaviors.create!(curve_type: 'inflex', behavior: 'generic')
```

antw:
> Yes. I added a visible property to the technologies table which takes care of that.

So you did; sorry!

app/models/installed_technology.rb
antw: Looks like my original code also wrongly used `default` rather than `generic`. My bad. :disappointed:

grdw: Makes more sense. I'm also a bit clueless when I came up with these lines of code... 
Thanks for pointing them out. I'll look into it.

grdw: @antw I changed these lines of code. Didn't add the behaviors yet for `base_load_edsn`.

grdw:
All the changes are done thusfar. If you could roll your eyes over this pull request one more time.

antw:
I'm getting this whenever I try to import or edit:

```
NoMethodError - undefined method `profile_table_options_for_households' for #<#<Class:0x007fbd51e57ae8>:0x007fbd578e3480>:
  app/views/testing_grounds/_profile_table.html.haml:20:in `block in _app_views_testing_grounds__profile_table_html_haml__3220321190808077264_70225559146760'
  activerecord (4.1.8) lib/active_record/relation/delegation.rb:46:in `each'
  app/views/testing_grounds/_profile_table.html.haml:14:in `_app_views_testing_grounds__profile_table_html_haml__3220321190808077264_70225559146760'
  ...
```

Looks like [an old version of the method](https://github.com/quintel/etmoses/commit/5c020cccf2ed50516eafda8a84ef990997972c0c#diff-bd8bcc0fab6bc697abd6b062e30b4bdf) was [removed in a rebase](https://github.com/quintel/etmoses/commit/9332452e65df6fda22c6c55bcf8db882fa741825#diff-bd8bcc0fab6bc697abd6b062e30b4bdf). I tried to bring back the old version of the method, but it relied on `LoadProfile#is_edsn?` which doesn't exist anymore.

grdw:
That helper method was moved to another helper, but changed in the previous helper, which caused a conflict. I removed the method later on because it's no longer used.

antw:
> If you could roll your eyes over this pull request one more time.

Looks good, merged! No eye-rolling involved. :smiley:

PR #332, opened 2015-08-28
Issues/#285 defaults for technologies

### LINKED ISSUE #285: All technologies should get appropriate default values
This table summarises the debts for the assignment of default values from the different technologies in ETMoses.  Boldface capital words mean that this still has to be done.

|Technology| Still needs capacity|Capacity should be fixed|Still needs volume|Volume should be fixed|Comments|
|---|---|---|---|---|---|
|HP space heating (air)| no | **YES** | **YES** | **YES** | (see https://github.com/quintel/etmoses/issues/270)|
|HP space heating (ground)| **YES** | **YES** | **YES** | **YES** | (see https://github.com/quintel/etmoses/issues/270)|
|HP hot water (air)| no | **YES** | **YES** | **YES** | (see https://github.com/quintel/etmoses/issues/270)|
|HP hot water (ground)| **YES** | **YES** | **YES** | **YES** | (see https://github.com/quintel/etmoses/issues/270)|
|P2G| **YES** | no | no | no ||
|P2H| **YES** | **YES** | **YES** | **YES** ||

grdw:
Fixes #285

### DIFF
--- a/app/assets/javascripts/profile_select_boxes.js
+++ b/app/assets/javascripts/profile_select_boxes.js
@@ -1,10 +1,12 @@
 var ProfileSelectBoxes = (function(){
+  var etmDefaults;
   var defaultCells = ["Capacity", "Volume", "Demand"];
 
   ProfileSelectBoxes.prototype = {
     add: function(){
       addProfileSelectBoxes();
       addChangeListenerToNameBox();
+      addChangeListenerToProfileBox();
     },
 
     update: function(){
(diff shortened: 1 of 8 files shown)


antw:
I think this PR depends on #331 (there are EDSN-related changed in installed_technology.rb); after #331 is merged, I'll also merge this.

antw:
I rebased against master to account for the changes there, but there are some quirks to the form (which may be intentional). @ChaelKruip @grdw can you tell me if this is intended behaviour?

For these examples, I (unrealistically) added a `default_volume` of 10.0 to the solar PV technology.

* Default values do not appear on the "import" form when it first loads; I have to change the profile to get the default value to show up. The -1.5 *capacity* does show up: this value is imported from ETEngine.

* If I change the selected profile, the -1.5 capacity now disappears, and the volume of 10.0 shows up.

My expectation is that:

1. Default values should be shown in the table when it first loads.

2. If a field is editable, an existing value should not be automatically deleted or changed. Perhaps the value *can* be altered if the user has made no changes (current value == default value) but this needs to account for values which are imported from ETEngine which should also not be deleted.

3. If the value of a field is a function of the selected profile (e.g. an EV profile which has a hard-coded capacity of 3.7kW) then the field should not be editable by the user (set `disabled="disabled"`).

ChaelKruip:
> can you tell me if this is intended behaviour?

No. Your expectations are my expectations.

grdw:
> Default values should be shown in the table when it first loads.

So ignoring the values that are imported from ETEngine?

ChaelKruip:
> So ignoring the values that are imported from ETEngine?

No. I think the values from ETEngine are a 'special kind' of defaults. There are, however also default values that are not present in ETE such as 'volume'.

In [this comment](https://github.com/quintel/etmoses/issues/285#issue-102754112) I have described which default values should be 'fixed' (not editable by the user but depending on the profile or the technology).

antw:
> So ignoring the values that are imported from ETEngine?

User-provided value > ETEngine value > default value.

antw:
To put it in more abstract terms: the user experience should be consistent. Here's what would *currently* happen:

* The page loads and my technology has selected "profile_one", a capacity of -1.5 and no volume.
* I change the profile to "profile_two"; my technology now has no capacity and a volume of 10.0.
* I change the profile back to "profile_one", but my technology has no capacity and a volume of 10.0.

The user experience is *inconsistent* because, despite reverting my changes, the technology row has different values (capacity and volume) than when the page first loaded.

But, if we can apply the rule above – user value > ETE value > profile default > technology default – then hopefully the UX becomes consistent when a user makes, and reverts, changes.

grdw:
I've made changes according to:

> User-provided value > ETEngine value > default value.

If you would be so kind to check again.

antw:
Looks good, thanks!

Issue #308, opened 2015-08-25
The market model table needs a re-style


ChaelKruip:
Currently it has the following columns:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/9467778/4efe8bd4-4b3e-11e5-9830-38032a48e54d.png)
This should become:

|From stakeholder|To stakeholder|Measure|Applied to stakeholder|Tariff (EUR/unit)|
|---|---|---|---|---|
|Customer|Supplier|kWh consumed|Customer|0.22|

As described in https://github.com/quintel/etmoses/issues/219

ChaelKruip:
Including @antw as this ties in with the way the measures work ('measure' and 'applied to').

grdw:
What's the purpose of the 'Applied to stakeholder' field? Is that a completly different field than `from` and `to`?

ChaelKruip:
> What's the purpose of the 'Applied to stakeholder' field? Is that a completly different field than from and to?

Yes. That is the stakeholder that the measure is applied to. So, the customer (from) might pay the supplier (to) for the number of kWh produced (measure) by the producer (stakeholder that the measure is applied to).

grdw:
![screen shot 2015-09-01 at 10 09 48 am](https://cloud.githubusercontent.com/assets/2676542/9599133/99cbe8a2-5091-11e5-8f31-2ddd80c5ef86.png)

Does that imply that the edit and new screen also need another dropdown box, selecting to which stakeholder (from or to) the measure applies to?

PR #338, opened 2015-09-01
Issues/#308 market model restyle

grdw:
Fixes: #308

### DIFF
--- a/app/views/market_models/show.html.haml
+++ b/app/views/market_models/show.html.haml
@@ -1,17 +1,17 @@
 %h1
   = @market_model.name
-%table.interactions.table
+%table.interactions.table.table-bordered
   %tr
-    %th Stakeholder from
-    %th Stakeholder To
+    %th Payer
+    %th Payee
     %th Measure
-    %th (not used)
-    %th Tariff Type
-    %th Tariff Price
+    %th Applied to stakeholder
+    %th Tariff (EUR/unit)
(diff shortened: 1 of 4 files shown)


app/views/market_models/show.html.haml
```diff
@@ -1,17 +1,16 @@
 %h1
   = @market_model.name
-%table.interactions.table
+%table.interactions.table.table-bordered
   %tr
-    %th Stakeholder from
-    %th Stakeholder To
+    %th From stakeholder
+    %th To stakeholder
```
antw: @ChaelKruip @grdw Would it be more meaningful to rename "From stakeholder" and "To stakeholder" to "Payer" and "Payee" respectively? I'm struggling to think of a situation where "from" isn't paying "to", and these terms might be a bit easier to understand to novice users.

ChaelKruip: I like that! :+1:

antw:
To add to the comment about "payer" and "payee", the market model *form* also needs to be updated to remove the "Measure" column, and rename "Foundation" to "Measure".

screen shot 2015-09-01 at 12 46 12

The table on "market_models/show" looks good now.

grdw:
screen shot 2015-09-07 at 3 42 21 pm

Like this?

antw:
:+1:

Issue #337, opened 2015-08-31
Show LES interface should be restructured


dennisschoenmakers:
As discussed in the meeting with @grdw and @ChaelKruip 

![img_1241](https://cloud.githubusercontent.com/assets/104830/9581907/6fbdfcce-5002-11e5-9214-72aaff5d6396.JPG)

grdw:
![screen shot 2015-09-02 at 2 24 19 pm](https://cloud.githubusercontent.com/assets/2676542/9631019/544a8e76-517e-11e5-90c7-9b71df9c9128.png)

Including @ChaelKruip @antw @dennisschoenmakers 

If the topology and load chart are next to each other; this is the result. 
I want to hear your thoughts.

antw:
We used to have the topology top-down – instead of left-to-right – and abandoned it because it was difficult to show complex networks. *However* left-to-right [still has that problem with very large networks](https://cloud.githubusercontent.com/assets/2676542/9575723/1417fac6-4fd2-11e5-95ce-17ccedaec094.png) and I actually prefer top-down. So I think that's a win.

I also like the chart being next to the topology, as that results in less scrolling.

I would suggest "Pick a strategy" becomes "Choose network features..." or something similar; "strategy" is a bit too technical.

Overall I think this is a big improvement. :+1:

dennisschoenmakers:
I like the charts next to each other, absolutely. :+1: 

It should be even more **clear** which node is selected, I think the current underline is too subtle.

grdw:
![screen shot 2015-09-03 at 10 10 09 am](https://cloud.githubusercontent.com/assets/2676542/9655152/6eb8594e-522f-11e5-9e97-18ce81e44a54.png)

How about some opacity?

ChaelKruip:
> How about some opacity?

I like!

grdw:
Full-preview

![screen shot 2015-09-03 at 3 44 03 pm](https://cloud.githubusercontent.com/assets/2676542/9659806/aee0fecc-5252-11e5-8a00-ea800ee2b12f.png)

ChaelKruip:
Very nice! :+1: :+1:

antw:
> "When clicking a node, load will appear"

* Turn this into a call-for-action with something like "Click on a network component to view its load and attached technologies"; this encourages the user to do something (explore and experiment). "Network component" is likely more meaningful to them than "node", which is a comp-sci term.

[...]

PR #351, opened 2015-09-07
Issues/#337 les interface restructuring

grdw:
Fixes: #337 
Fixes: #352


antw:
I made a couple of small visual tweaks:

* Lightened the background on selected strategies; the default dark blue was a bit overwhelming, and clashes with the OS X 10.10 checkbox colour,

* Fixed the padding in the multi-select (the bootstrap defaults were overriding the multi-select CSS styles),

* Reduced the size of the helper text beneath the solar PV slider (prevents the drop-down from growing in width when solar PV capping is enabled),

* Made the multi-select always read "Pick a strategy (__ selected)". The default behaviour was to replace the button text with the names of the chosen options, but that didn't feel right in practice since our options are so wordy.

* Moved the "Edit" button to the right, away from the title (and added an icon to draw attention to it).

screen shot 2015-09-08 at 12 32 57

What do you think?

Two other things stand out to me:

* boostrap-multiselect has a `includeSelectAllOption` which would be a good usability improvement for when you want to turn on all strategies at once. I tried to add this, but it didn't work as expected in our app. It would select all options, but the "Select all" checkbox would remain unticked, and I couldn't click it again to deselect all the strategies. Not a big loss if we can't use it, but it would be nice.

* [Currency formatting in the business case table](https://github.com/quintel/etmoses/pull/351/files#diff-6c6f6d2aeeec8f74c8dd28011f1bd5e4R5) seems to be in the decimal-comma style ("€5.000,00") rather than decimal-point ("€5,000.00"). Was this intentional? I only ask because other numbers in the app – and currency figures in other applications (ETM) – use the decimal-point style.

  My understanding is decimal-point is standard in scientific use, but perhaps @ChaelKruip can shed some more light on this. Maybe decimal-comma *is* used in non-English papers? I'm fine with using decimal-comma if we're focusing on an EU-only audience, but I think it has to be applied consistently to all numbers, not just one table.

Otherwise, I'm really happy with this and can't wait to get it merged. :+1:

grdw:
> What do you think?

That looks good! :+1: 

> Two other things stand out to me:

> bootstrap-multiselect has a includeSelectAllOption 

Yes, I'll fiddle with it. 
 
> Currency formatting in the business case table seems to be in the decimal-comma style ("€5.000,00") rather than decimal-point ("€5,000.00"). Was this intentional? 

No, it isn't, but if decimal-point is the default I'll change it to decimal-point.

ChaelKruip:
Excellent work guys! This looks sooo much better than before. 

> No, it isn't, but if decimal-point is the default I'll change it to decimal-point.

Decimal point makes more sense to me as well as it is consistent with the rest of the ETM. Well spotted @antw!

grdw:
> Yes, I'll fiddle with it. 

I fiddeled, but I'm guessing this is the type of "I'll fix this in 5 minutes, 8 days later it still isn't solved" -type of issue.

So I'll make an issue of it for a later moment in time. I have no idea why it isn't working properly.

antw:
> I fiddeled with the includeSelectAllOption, but I'm guessing this is the type of "I'll fix this in 5 minutes, 8 days later it still isn't solved" -type of issue.

I know the feeling. :anguished:

One merge coming up!

grdw:
> I made a couple of small visual tweaks:

@antw  are these visual tweaks also merged or?

antw:
Oops. I didn't push them to your branch. :blush: 

I'll fix it now, sorry!

Issue #376, opened 2015-09-11
Business case calculation with strategies


grdw:
I currently made it possible to make the business case calculation with strategies.

- Save strategies with testing ground

Issue #365, opened 2015-09-10
Business cases should be comparable to one from another LES


ChaelKruip:
This used to be possible but now the button has disappeared.

Issue #364, opened 2015-09-10
It should not be possible to view a LES containing a private topology that you don't own


ChaelKruip:
See [http://beta.moses.et-model.com/testing_grounds/243](http://beta.moses.et-model.com/testing_grounds/243)

grdw:
@dennisschoenmakers came up with a nice solution. All the testing grounds that use a private topology or market model for that matter, are by default private.

And I would like to add that you cannot change your privacy settings of a testing ground unless both the market model and the topology are public.

ChaelKruip:
:+1:

grdw:
This applies only to future LES's for now. I should also run a migration to make all current LES's this way.

grdw:
Fixed.

Issue #359, opened 2015-09-09
Calculate business case in background job


grdw:
In the current 'ajax' #358 situation. The market model and topology can be editied, but the business case has to change accordingly. Recalculating the business case costs a lot of time. 

@ChaelKruip Chael suggested to move the freeform rules to the market model. Which is fine with me.
This way we can remove the business case tab from the edit screen and calculate once the user is on the LES-page.

antw:
To take this a bit further... it would be nice to also do the network calculation in a background job. Currently the LES loads with a "please wait..." message while an XHR request is sent to Moses for the load data.

This XHR request hits `testing_grounds/:id/data` where the loads are calculated in the Rails process. This is an *okay* solution provided we have enough processes, but moving this to a background worker – and having the front-end poll to see when the calculation is finished – would be a better solution.

Something like:

```
GET testing_grounds/1/data
{ pending: true }

... time passes ...

GET testing_grounds/1/data
{ graph: {,...}, technologies: {,...} }
```

> Recalculating the business case costs a lot of time.

Mostly because it has to calculate the network first. It's going to get even worse when the flexibility measures get hooked up (#330). :cry: 

We shall really have to get some form of caching sorted for the network once the major issues are out of the way.

grdw:
> To take this a bit further... it would be nice to also do the network calculation in a background job. 

Do you have any preference for a background processor? 

Resque / DelayedJob / Sidekiq / ActiveJob ?

grdw:
> and having the front-end poll 

This is in my experience still a tricky one. Without using `setInterval` how could this be done nicely?

antw:
> Do you have any preference for a background processor? Resque / DelayedJob / Sidekiq / ActiveJob ?

*Maybe* Resque? In truth I don't have a huge amount of experience with any of them; so if you have a preference, I'm happy with whatever you choose.

> Without using setInterval how could this be done nicely?

WebSocket, but that feels like overkill. `setInterval`/`setTimeout` is probably how I'd have done it:

```javascript
function fetchLESData(callback) {
  $.ajax(...).success(function(data) {
    if (data.pending) {
      // Maybe return something other than 200 OK?
      window.setTimeout(function() { fetchLESData(callback) }, 1000);
    } else {
      callback(data);
    }
  })
}
```

PR #377, opened 2015-09-11
Adding delayed job

grdw:
Fixes: #359 
Fixes: #364 
Fixes: #365 
Fixes: #376 

Start delayed job with: `rake jobs:work`


grdw:
@antw I fixed some of your comments. The only thing that is still missing is the "`beforeSend` and `pending: true`"-part because I didn't really understand what you meant.

antw:
> After some reading. progress and done are the functions used for $.Deferred.
>
> Thanks for the suggestion, this is really awesome.

You're welcome! I know you're busy, so I didn't expect you to actually make this change; it was just a suggestion for future callback-based code. Impressive that you did it anyway. :+1:

> The only thing that is still missing, is the "beforeSend and pending: true"-part because I didn't really understand what you meant.

:laughing: 

I meant that jQuery's $.ajax() function has a `beforeSend` callback which is run before a request is made, and a `success` callback for when it finishes successful. Maybe it would be nice for `Poller` to use the same callback names instead of `pending` and `final_success`. Doesn't matter now though.

I will have to add monitoring for the delayed job process to our Ansible repo (likely tomorrow). I'll merge this as soon as I've completed that.

Thanks again!

antw:
I merged without the process monitoring; I'm short on time today and don't want to hold up your other work by delaying the merge any longer. (Also, I thought I could kill two birds with one stone by also setting up monitoring for the [Optimizer](https://github.com/quintel/opt) queue... then realised that uses Resque. :confounded:) I'll make sure it happens before we deploy to production.

I have deployed this to staging. Thanks @grdw!

PR #391, opened 2015-09-15
Upgrade Airbrake

dennisschoenmakers:
I would like to have deployment integration.

This script should generate the `config/initializer/airbrake.rb` file according to the [manual](https://github.com/airbrake/airbrake): 

```
bundle exec rails generate airbrake --api-key THE_ETMOSES_AIRBRAKE_KEY
```

But it complains here:

```
Don't know how to build task 'airbrake:test'
```

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -50,5 +50,6 @@ namespace :deploy do
   end
 
   after 'deploy:compile_assets', 'paperclip:build_missing_styles'
+  after 'deploy:finished', 'airbrake:deploy'
   after :publishing, :restart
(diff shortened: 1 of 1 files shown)


twilkinson573:
@dennisschoenmakers if you add the following to `secrets.yml` then the airbrake deployment should be working: 
```ruby
staging:
  airbrake_key: <etmoses_airbrake_key>
```

dennisschoenmakers:
Currently, the secret hash is already there on beta.moses.et-model.com. Merging.

Issue #236, opened 2015-07-21
If a measure does not exist for the selected topology, the user should be warned


ChaelKruip:
I think that there should be some feedback after the user clicks the  'calculate business cases'  button (which we do not yet have). 

The warnings/error messages should be quite specific because things can go wrong in several locations:

* The topology does not contain the information necessary for a measure:
  * Depreciation costs rely on **investment_cost, economic_lifetime, yearly_o_and_m_cost**
  * If stakeholders are not included, the measures do not know who to apply to
  * If capacities and 'units' are not specified for end-points, the measure for connections fails
* The stakeholder is defined in the topology but not in the Market Model (this does not need to be a problem, but the user should probably be warned)

grdw:
> Depreciation costs rely on investment_cost, economic_lifetime, yearly_o_and_m_cost

How does this work exactly? 

> If stakeholders are not included, the measures do not know who to apply to

This means that my market model measures contain `stakeholders` that aren't in the topology. And the calculation should stop and throw an error?

> The stakeholder is defined in the topology but not in the Market Model (this does not need to be a problem, but the user should probably be warned)

And the other way around.

ChaelKruip:
> How does this work exactly?

This is the 'transaction' from the stakeholder to itself in the matrix (see 'investments' cells in https://github.com/quintel/etmoses/issues/213). It is the sum of:
* Depreciation cost: investment / lifetime
* O&M cost per year (given by yearly_o_and_m_cost)

ChaelKruip:
> This means that my market model measures contain stakeholders that aren't in the topology. And the calculation should stop and throw an error?

Not necessarily. Stakeholders do not have to be in the topology to send or receive money. But they do have to be in the topology to have a **measure apply to them**.

ChaelKruip:
> And the other way around.

Not really. The Government, Supplier, Aggregator and Producer are usually not physically associated with any nodes of the topology. The user does not need to be warned about this.

grdw:
We talked about this last friday I remember. The thing was that if 

![screen shot 2015-09-14 at 10 33 57 am](https://cloud.githubusercontent.com/assets/2676542/9845559/225fbdbc-5acc-11e5-9bc6-cff99fa298fe.png)

One of these does not appear in the topology the user should get a warning. 

If I understand correctly.

ChaelKruip:
> If I understand correctly.

[...]

PR #399, opened 2015-09-16
Issues/#236 validate market model topology

grdw:
Resolves: #236


antw:
Looks good! I like that it validates on-the-fly whenever I change the topology/market-model fields on the import form.

PR #436, opened 2015-09-24
Issues/calculating default costs in business case

grdw:


dennisschoenmakers:
Holler when you merge and deploy this @antw! I want to see this :smile:

app/models/market/initial_costs.rb
```diff
@@ -0,0 +1,53 @@
...
+    def technology_costs
+      group_sum(technology_nodes) do |node|
+        node.techs.map(&:installed).sum do |tech|
+          tech.initial_investment / tech.technical_lifetime
+        end
+      end
+    end
+
+    def group_sum(nodes)
+      Hash[nodes.group_by{|node| node.get(:stakeholder) }.map do |_, nodes|
```
antw: FYI: Naming a variable `_` is [meant to indicate that the variable is not used](https://github.com/bbatsov/ruby-style-guide#underscore-unused-vars). You can also give it a name, but prefixed with `_` [such as here](https://github.com/quintel/etmoses/blob/ee3522363053313835eb391acc095218412d6298/app/models/network/technologies/generic.rb#L60). In this case, the variable *is* used (as the first element of the array on line 33), so it should have a name.

See [Ruby feature #6693](https://bugs.ruby-lang.org/issues/6693).

antw:
I pushed a couple of small changes (notably so that calculation could still proceed if a node had *some* technologies with costs assigned, and some without). However, there's a spec failure which started with c0d188a:

```
  1) Finance::BusinessCaseComparator compares two business cases
     Failure/Error: expect(compare).to eq([[-4, -1, 0, -3], [0, -3, 0, 3]]);

       expected: [[-4, -1, 0, -3], [0, -3, 0, 3]]
            got: [[-3, -1, 0, -2], [0, -3, 0, 3]]

       (compared using ==)
     # ./spec/services/finance/business_case_comparator_spec.rb:25:in `block (2 levels) in <top (required)>'
     # ./spec/support/fixtures.rb:29:in `block (2 levels) in included'
     # ./config/application.rb:75:in `with_data_dir'
     # ./spec/support/fixtures.rb:29:in `block in included'
```

Otherwise, the PR looks good and works well. :+1:

grdw:
@antw that's right. The reason why it's failing it's because the initial investment doesn't need to be included in the total `incoming`.

antw:
Notifying @dennisschoenmakers as requested. This has been merged and deployed.

PR #444, opened 2015-09-24
Store EditableTable events in the instance

antw:
Moved `changeListener` and `changeData` callbacks to be stored within the EditableTable instance, instead of being global within the anonymous closure. The old global variables would have their values overwritten each time a new EditableTable was created.

**Why this is needed?**

Previously, if muliple EditableTable instances were created (on the LES edit page) the callback event from the most-recently created table would replace the event on the older ones. In particular, it meant we couldn't add or remove rows to the market model table because the callback event from the *technology profile table* was being run instead. The correct callback would be run, apparently, only if you edit a value in the market model table prior to submitting the form.

**Steps to reproduce**

1. Visit the "Edit LES" page
2. Click the "Market model" tab
3. Remove a row
4. Without focusing any of the form fields in the table, press "Save"
5. Refresh the page, and note that the deleted row has reappeared.

@grdw Could you review this and let me know what you think? I'm not sure what the reason was for the global `changeListener` and `changeData` variables, but as this isn't a technique I've seen before for creating JS classes, I don't want to make this change to `master` without you reviewing it first.

### DIFF
--- a/app/assets/javascripts/tables/editable_table.js
+++ b/app/assets/javascripts/tables/editable_table.js
@@ -1,12 +1,10 @@
 var EditableTable = (function(){
-  var selector, changeData, changeListener;
-
   EditableTable.prototype = {
     append: function(_changeListener, _changeData){
-      changeListener = (_changeListener || function(){});
-      changeData = (_changeData || function(){});
+      this.changeListener = (_changeListener || function(){});
+      this.changeData = (_changeData || function(){});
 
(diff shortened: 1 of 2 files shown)


grdw:
I'm ignoring semaphoreci here.

antw:
Thanks! Semaphore can indeed be ignored as it hasn't been set up yet.

Merged by hand.

Issue #477, opened 2015-09-26
Green-line loads could be fetched in parallel with blue-line loads


antw:
When an LES page loads, it fetches the all-strategies-off (blue line) loads. Once received, the JavaScript then asks the server for the green-line loads.

Instead of executing these requests in series, both requests could be sent simultaneously. This would improve the responsiveness of the "show LES" page.

@grdw Do you think this is this doable with the current "show LES" page?

ChaelKruip:
Good catch @antw and I like your suggestion!

grdw:
This is not to big of change. I'll look into it!

Issue #476, opened 2015-09-26
Clicking a node prior to receiving the green-line loads raises a JS error


antw:
When an LES page first loads, it fetches the "all-strategies-off" (blue line) loads. When finished, the network diagram shows up. At this point the JS is still waiting for the "all-strategies-on" (green line) loads. If you click on a node in the network diagram prior to the green-line loads being received, a JavaScript error is raised:

![screen shot 2015-09-26 at 20 00 09](https://cloud.githubusercontent.com/assets/4383/10119269/3871e654-6489-11e5-93a7-68ca091570d9.png)

PR #485, opened 2015-09-28
Load normal loads + strategy loads in a single go

grdw:
Fixes: #476 
Fixes: #477 
Fixes: #487


antw:
Good stuff @grdw; the difference is really noticable!

**Before**

screen shot 2015-09-29 at 12 26 46

**After**

screen shot 2015-09-29 at 12 25 50

ChaelKruip:
Nice!

Issue #492, opened 2015-09-29
User sign up is broken because of activation email not working/not communicated properly to the user


dennisschoenmakers:
Now it's really confusing. If you sign up, you get redirected to the login page saying 'you need to sign in/ sign up'. If you type in your credentials again you get the cryptic:

![image](https://cloud.githubusercontent.com/assets/104830/10164761/abedc4f2-66bb-11e5-9b5f-e735ab083903.png)

* get rid of the activation requirement
* sign in when you sing up immediately.

cc @ChaelKruip

ChaelKruip:
:+1:

grdw:
> get rid of the activation requirement 

Ok.

grdw:
> sign in when you sing up immediately

Which song should they sing? :smile:

dennisschoenmakers:
Knocking on heavens door?

PR #497, opened 2015-09-30
Updating values directly from ETE

grdw:
Please consider that #492 needs to merged first.

### DIFF
--- a/app/models/import/variable_om_costs_per_full_load_hour_attribute.rb
+++ b/app/models/import/variable_om_costs_per_full_load_hour_attribute.rb
@@ -1,0 +1,5 @@
+class Import
+  VariableOmCostsPerFullLoadHourAttribute =
+    Attribute.new('om_costs_per_full_load_hour', 'variable_operation_and_maintenance_costs_per_full_load_hour')
+end
+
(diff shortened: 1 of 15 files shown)


ChaelKruip:
The title and the content of this PR do not seem to belong together @grdw

grdw:
The first two commits belong to PR #492

app/models/import/variable_om_costs_per_full_load_hour_attribute.rb
```diff
@@ -0,0 +1,5 @@
+class Import
+  VariableOmCostsPerFullLoadHourAttribute =
+    Attribute.new('fixed_om_costs_per_year', 'variable_operation_and_maintenance_costs_per_full_load_hour')
```
antw: I think the `local_name` argument ("fixed_om_costs_per_year") may be wrong. Was it meant to be "om_costs_per_full_load_hour"?

grdw: Yes. It should be. Moment please.

antw:
Had to merge manually due to a conflict, but it looks good and works nicely. Thanks! :+1:

Issue #514, opened 2015-10-07
Solar slider should show up when capping is selected


ChaelKruip:
Now it disappears!
![etmoses](https://cloud.githubusercontent.com/assets/1303760/10336846/e61abd74-6cfb-11e5-94fb-9b2e6a9dc523.png)

Issue #508, opened 2015-10-05
An existing LES should not need to re-calculate when viewed


grdw:
This is a general issue.

It takes a long time to load a LES while the calculation is already done. 

We could save the calculation to the database in order to save time. Don't quiet know what the limitations are database wise and when it comes to `$.ajax`. So I'll have to look into that first.

- [ ] Researching options for storing large amounts of data to the database
- [ ] Researching limitations of browsers

grdw:
*If* we were to store it in the database. The thing would be, to store the 35040 points per end-point in `MEDIUMBLOB` type columns. Thing is though, if you have 200 end-points - it will still be slow.

The other options are: 

- Redis (using `ENDPOINT_KEY_NAME => [ 35040 values ]`)
- Writing to file (this seems really slow mentally but maybe in practice it's really fast)

antw:
I put some thought into this recently, but I will start by saying these ideas are off the top of my head. I have some experience of building simple caching systems, but this task will be tricky because of the amount of data we need to store at once, and the way in which we need to access it.

### Requirements

* We need to store at least two versions of the computed network: the “blue-line” and “green-line” loads.

* The blue-line and green-line loads should be separate, since the blue-line load rarely needs to be recalculated. The green-line load, however, may be frequently recalculated as a user plays with the “technology behaviour” settings in the front-end.

* The cache should be available with a Ruby API to fetch the data. For example:

  ```ruby
  cache = NetworkCache.for(TestingGround.first, strategies)
  cache.node('HV Network’).load
  ```

  This will allow us to extend the API at a later date with the nice-to-haves.

* Ideally the API should load the minimal amount of data needed to fulfil the request. The previous example should load the “HV Network” load only. This probably means storing each node load in a separate file/cache-key/etc.

* It should be possible to disable the cache for development and testing purposes, and get a fully-recalculated network each time.

#### Nice-to-have

Not needed in the initial version, but considerations for the future:

* Store the loads of individual *technologies* to be shown in a chart.

* Store the “storage” amounts of individual technologies, so that we may visualise how much is stored in a battery or heat-tank at each moment in time.

### Cache Expiry

* Green-line loads need to be expired and recalculated when a user changes the technology behaviour settings.

* Blue- and green-line loads need to be expired when the user changes the technology mix.

* Editing a topology should expire all caches for all LES’ using that topology.

* **All** caches should be expired when performing a deploy to staging or production.

### Storage

[...]

PR #518, opened 2015-10-08
Issues/#508 storing existing load calculation

grdw:
Implements a basic caching system which is the following:

- Caches the load calculation
- Clears the cache once the strategies are changed or when the topology / market model / technologies are in any way changed

This merge request also contains the following:
- Increase speed for removing the green line
- Only submit forms that have been edited when clicking 'Save all' - instead of submitting each form.
- Fixes: #514, #508


grdw:
@antw I fixed your comments :+1:

antw:
The main issue to be solved is the way caching bleeds across several layers of the application. Discussed in "[Transparent caching](https://gist.github.com/antw/11575308b15db05b75de#transparent-caching)", neither the front-end nor the network calculation *process* should be involved in caching. This means:

1. The `clear_cache` parameter should not be needed by the JS in order to expire the cache.
2. The `Network` and `Calculation` classes [should not be writing the cache](https://github.com/quintel/etmoses/commit/6a703fc23cebbbdf9125a42dc884e799e4488463#diff-27a7aedba59d5f302ac9ff620caf0273R22).

#### Suggestions:

1. Have the Rails controllers or [models expire the cache](https://gist.github.com/antw/11575308b15db05b75de#expiring-when-updated) instead of using a `clear_cache` parameter.
2. Instead of using `NetworkCache::Writer` in `Calculation::PullConsumption`, write to the cache *after* calling `TestingGround#to_calculated_graph`.

  For example:
  
  ```ruby
  class TestingGroundController
    def data
      network = if NetworkCache.present?(testing_ground)
        # The network has a cache stored on disk. Load network from cache.
        NetworkCache.write(...)
      else
        # Network is not cached. Calculate it, then cache the results.
        network = testing_ground.to_calculated_graph
        
        # Write cache, then...
        NetworkCache.write(network, ...)
        
        network
      end
      
      respond_with network
    end
  end
  ```
  
  This could probably be simplified as:
  
  ```ruby
  class TestingGroundController
    def data
      respond_with(NetworkCache.fetch(testing_ground, opts) do
        testing_ground.to_calculated_graph(opts)
      end)
    end
  end
  
  module NetworkCache
    module_function
    
    # Reads the cache if present. If not (read returned nil) the block is executed
    # and the cache is written.
    def fetch(testing_ground, opts = {}, &block)
      read(testing_ground, opts) || write(testing_ground, opts, &block)
    end
    
    def read(testing_ground, opts = {})
      if present?(testing_ground, opts)
        # Read from cache.
      end
    end
    
    def write(testing_ground, opts, &block)
      # Remove existing cache if present, then...
      Writer.from(testing_ground, opts).write!(block.call)
    end
  end
  ```

grdw:
> Discussed in "Transparent caching"

It's not really much of discussion, but I get the point. Thanks for explaining. :+1: 

Will change this.

antw:
> It's not really much of discussion

discuss
* talk about (something) with a person or people.
* *talk or write about (a topic) in detail, taking into account different issues or ideas.*

I was going for the second meaning. :smiley:

> Thanks for explaining. :+1: Will change this.

The rest of the ideas in document can wait, at this stage they're "nice-to-have" rather than essential. Getting caching out of the front-end and `Network`/`Calculation` is the only thing we *need* to do, as I could see that causing us problems whenever we make changes to either. If that can be solved, we'll merge this and can revisit at a later date.

grdw:
@antw I managed to remove the caching from the load calculation and from the front end.

That's what I managed to do. What I didn't manage to do yet is the refactoring of the `Reader` and `Writer`. Like what you said, the `Reader` only contains one usable line.

Before I'm going to do that I want you to look at the current code and sort of say whether the direction this is going in, is any good.

antw:
> Before I'm going to do that I want you to look at the current code and sort of say whether the direction this is going in, is any good.

Looks like a *big* improvement. Refactoring `Reader` and `Writer` isn't strictly necessary, but it would be nice *if* you have the time available to do so. If not, I can start doing some testing as soon as you want.

grdw:
> but it would be nice if you have the time available to do so

That's the issue sort of. I was supposed to already start with the importing of Gas technologies. Which is now a little delayed. :sweat_smile:  I'll leave this for now and see if on the train ride home I can make some refactoring thingies.

grdw:
I did manage to refactor some things. Still would like to read up on the whole facade pattern (or any other cool patterns) before actually using it. Otherwise I'll just be copying and pasting your suggested code, which is something I won't learn that much from.

antw:
This looks pretty good to me. I did manage to find a way to break it, however: if you edit a topology on the "edit topology" page (`TopologiesController#edit`), the cache for affected networks doesn't get cleared.

grdw:
> if you edit a topology on the "edit topology" page (TopologiesController#edit), the cache for affected networks doesn't get cleared.

Yes. I might have a question about that, now that you mention it.

The `TestingGround::Cloner` class also destroys the cache. But would that be - according to [Transparent Caching](https://gist.github.com/antw/11575308b15db05b75de#transparent-caching) - a 
good practise? 

The easy way to Rome right now would be, is to expand the `update` method in the `TopologiesController`

````ruby
def update
  testing_ground = TestingGround.find(session[:testing_ground_id])
  NetworkCache::Destroyer.from(testing_ground).destroy_all
  
  @topology.update_attributes(topology_params)
  respond_with(@topology)
end
````

Which results in this heap of code. And I'm not sure if a controller should be aware of the caching?

**Comment by antw** [comment]
> The easy way to Rome right now would be, is to expand the update method in the TopologiesController

Pulling the testing ground out of the session isn't enough as there may be many using the same topology. I guess you'd have to:

```ruby
TestingGround.where(topology_id: topology.id).find_each do |tg|
  NetworkCache::Destroyer.from(tg).destroy_all
end
```

> The TestingGround::Cloner class also destroys the cache. But would that be - according to Transparent Caching - a good practise?

... I get the sense that you feel this is *not* a good practice; I agree. If time is limited, I'd say this is an acceptable compromise given that [something similar already happens in the TestingGround controlle](https://github.com/quintel/etmoses/blob/4644d1961c5165d15e71f03654bb3bb56159b80a/app/controllers/testing_grounds_controller.rb#L113).

The ideal solution is for the `NetworkCache` to determine if the cache is fresh (valid) or has expired based on the TestingGround it is given. This way, the Rails layer (the controllers, models, `Cloner`, etc) never need to know anything about the cache.

You'd need to save some metadata with the cache – namely the time at which the cache was written – and every time you read the cache check if any of the models have been updated. If their `updated_at` is *after* the cache was written, the cache is invalid, should be deleted and rewritten.

* For the "basic" graph, if the TestingGround or Topology `updated_at` is after the cache creation time, then the cache has expired.
* For the "feature" graph, if the TestingGround, Topology, or SelectedStrategy `updated_at` is after the cache creation time, then the cache has expired.

Perhaps that can probably accomplished somewhere in [`NetworkCache::Validator#valid?`](https://github.com/quintel/etmoses/blob/4644d1961c5165d15e71f03654bb3bb56159b80a/app/services/network_cache/validator.rb#L3-L5)

```ruby
def valid?
  cache_intact? && identical_strategies? && fresh?
end

private

def fresh?
  times = [
    testing_ground.updated_at,
    testing_ground.topology.updated_at,
    testing_ground.selected_strategy.updated_at
  ].max

  cache_created_at > times
end
```

--

Some other observerations:

* Why do `Fetcher`, `Destroyer`, `Validator`, etc, all inherit from `Cache`? I'm guessing it's because they all need [the same initialize method](https://github.com/quintel/etmoses/blob/4644d1961c5165d15e71f03654bb3bb56159b80a/app/services/network_cache/cache.rb#L7-L10)? If so, put it into a module and then `include` it into each class.

* Ideally, TestingGround::Calculator [should not write a SelectedStrategy](https://github.com/quintel/etmoses/blob/4644d1961c5165d15e71f03654bb3bb56159b80a/app/services/testing_ground/calculator.rb#L28) if one does not exist. This makes it not transparent as the presence of the cache has an effect on the database which would not happen if the cache were not there. (I think I'm realising that by "transparent" what I really mean is: "caching should have no side-effects on the rest of the application").

grdw:
> Perhaps that can probably accomplished somewhere in

Thanks! Implemented it by using the oldest `mtime` method (it grabs the oldest modification time for a cache file) and compare it against the updated at for the les and associated topology and market model.

> Why do Fetcher, Destroyer, Validator, etc, all inherit from Cache? I'm guessing it's because they all need the same initialize method? 

Yes and most of the private methods in `Cache`. But okay.

> Ideally, TestingGround::Calculator should not write a SelectedStrategy if one does not exist. 

How would that ideally look like? One global class which calls the `Calculator` and the `StrategyUpdater` separately?

antw:
> Yes and most of the private methods in `Cache`. But okay.

I'm not meaning to be difficult, I hope you know that. :scream:  

It's just that subclasses should have an "is-a" relationship with the superclass. `Dog` is an `Animal`. `Phone` is a `Device`. Any place you'd use an `Animal`, you should be able to substitute a `Dog` without anything breaking (this is "[polymorphism](https://en.wikipedia.org/wiki/Polymorphism_(computer_science)#Subtyping)" or "[duck typing](https://en.wikipedia.org/wiki/Duck_typing)").

But, as far as I could tell, a `Validator` isn't a `Cache`. In those cases, a shared module is clearer.

> How would that ideally look like? One global class which calls the Calculator and the StrategyUpdater separately?

I think if you write the SelectedStrategy (with all the strategies off?) when the TestingGround is created, the problem is solved. Updating the SelectedStrategy is probably the responsibility of the controller.

grdw:
> I'm not meaning to be difficult, I hope you know that. 

I know. I just wanted to know why. :smile:  Thanks for explaining!

Maybe my cold was talking here.

> I think if you write the SelectedStrategy (with all the strategies off?) when the TestingGround is created

Yes,

> Updating the SelectedStrategy is probably the responsibility of the controller.

Not necessarily would be my reaction. But I'll see if there's some nicer way.

grdw:
Not to self:
I'll keep the `around_filter`. When this is merged and when the background-calculation pull request will re-appear. I shall remove the `after_filter` and put it in the `success` method of the `Delayed::Job`.

grdw:
@antw is that 'okay' with you? I will create the other pull request as soon as this one is merged.

antw:
> I think if you write the SelectedStrategy (with all the strategies off?) when the TestingGround is created

I think something may have gone awry here. I tried to load a couple of testing grounds locally, and they all initially showed that no strategies were turned on. When I selected a strategy, the following exception gets raised (I guess because there's no `SelectedStrategy` for the LES?).

--

##### NoMethodError at /testing_grounds/351/data.json

> undefined method `update_attributes' for nil:NilClass

##### app/controllers/testing_grounds_controller.rb, line 205

``` ruby
  200       authorize @testing_ground
  201     end
  202   
  203     def update_strategies
  204       if params[:strategies]
> 205         @testing_ground.selected_strategy.update_attributes(strategy_params)
  206       end
  207     end
  208   
  209     # Internal: Before filter which loads models required for export-to-ETEngine
  210     # eactions.
```

--

I think it's because the app now expects a SelectedStrategy to be created when the TestingGround is ([here](https://github.com/quintel/etmoses/pull/518/files#diff-c2b7ca637ffbb6e651981cf3b6d89d2aR65)), but then breaks when an using an older LES which doesn't have a SelectedStrategy. This could be fixed by a migration to add a SelectedStrategy to each LES which doesn't already have one.

Otherwise, this pull request looks good. :+1:

grdw:
> This could be fixed by a migration to add a SelectedStrategy to each LES which doesn't already have one.

Yes. That's the case. I'll create a migration for it.

grdw:
@antw I added the pull request :smile:

grdw:
@antw I added the migration :smile:

antw:
Yes! Looks good! Thanks for the work; and for your patience with all of my requests. :laughing: 

Gonna merge this now. :+1:

grdw:
> and for your patience with all of my requests.

Haha, no problem. I learned a lot! So thanks.

PR #531, opened 2015-10-14
Topology-edit check if a child is removed

grdw:
- Throw an error if this happens
- Don't clone and don't update
- Ignore this validation when the graph is already incorrect

Fixes #531


grdw:
@ChaelKruip could you maybe validate this error message: 

https://github.com/quintel/etmoses/pull/531/files#diff-c6a4af8568e53f47391a0aacbe40d6e9R102

ChaelKruip:
> @ChaelKruip could you maybe validate this error message:

I think it is fine. Is that the type of validation you want?

grdw:
It's more to check if the error message is 'good'. If it's understandable to the user. I would say it is but maybe you have a different opinion about it.

app/models/topology.rb
```diff
@@ -92,4 +93,14 @@ def validate_stakeholders
       end
     end
   end
+
+  def validate_clone
+    return unless original.present?
+    return unless each_node
```
antw: What's the intent of this line? `each_node` [will return false if there are already errors present on the topology](https://github.com/quintel/etmoses/blob/b4ee9c19f6cf8622fa75ac543807b6128a4ca722/app/models/topology.rb#L34), but will returns a truthy value in all other situations.

If that's really what you want, I think it best to check explicitly for that instead of relying on the behaviour of `each_node`.

```ruby
def validate_clone
  return if original.blank? || errors[:graph].any?
  # ...
end
```

app/models/topology.rb
```diff
@@ -92,4 +93,14 @@ def validate_stakeholders
...
+
+    if each_node.count != original.each_node.count
```
antw: [Prefer `#length` or `#size` over `#count`](https://github.com/bbatsov/ruby-style-guide#count-vs-size). To be fair, it makes no different in this situation as `each_node` without args returns an `Enumerator`, but it's a good habit to be in.

app/models/topology.rb
grdw: This is currently fixed.

app/models/topology.rb
antw: I just realised `Enumerator` doesn't implement `#length`, so that's why you used `#count`. Sorry, my bad.

Issue #537, opened 2015-10-15
investment cost of nodes but no end-nodes


JeroenvdLogt:
I see in the business case table no investment cost of nodes which are no end-nodes.
you can look at Oefencase Etten-Leur SSU in the live server.

ChaelKruip:
@JeroenvdLogt please include a link to [your LES](http://moses.et-model.com/testing_grounds/341) (like this :smile:)

ChaelKruip:
I see that the `system operator` has no depreciation costs although they should be equal to:
```
400000 / 50 + 25 * 6000 / 30 + 25 * 7500 / 30 = 19350 EUR/year
```

Assigning to @grdw

grdw:
I already see what is happening here.

In some Topologies `technical_lifetime` is used to specify the 'duration' of a certain technology. In others the term `economic_lifetime` is used. The term to use for succesfully calculating the depreciation costs is `technical_lifetime`. There are no calculations that make use of the term `economic_lifetime`.

In order to not cause any future confusion about this term and it's usage, we could settle for a certain term. I don't feel a strong preference to either `economic_lifetime` or `technical_lifetime`. But maybe there are people with a stronger preference?

JeroenvdLogt:
If calculations are right it is not important which term we chose I think. For Regulation Affairs there is however a big difference between economic_lifetime and technical_lifetime. To give an example: there are some cables in dutch networks older than 70/80 years (technical_lifetime). But in order to proper benchmark between DSO's a economic_lifetime is choosen (based on statistics about lifetime of large population of assets). This economic_lifeime is used to eventually calculate total needed investments and to determine tariffs. 

So technical_lifetime can be seen as the actual lifetime in the field (of individual assets) and economic_lifetime as a statistic value saying something about the average life_time of such a type of assets. This average life_time is used in calculations for (re-)investments and tariffs.

I will discuss this with Hans and Bianco because in some usecases I see this same issue (difference of economic and technical lifetime) keep coming back.

ChaelKruip:
@JeroenvdLogt thanks for expanding on the subject. 

I suggest that for the moment we allow both to be used. The lifetime (whichever one) is only used to determine the depreciation costs at the moment. It does not really matter which is used as long as it is representative for how to spread the investment. 

Anyone opposed?

antw:
> I suggest that for the moment we allow both to be used.

:+1:

grdw:
[...]

Issue #530, opened 2015-10-14
When creating a LES, could we show a 'preview' of the topology


dennisschoenmakers:
When presenting ETMoses, I notice that people would like to preview the topology before it is chosen. It's also hard to change it later.

Could we improve on this? I have two ideas:

1. Show the collection of three default, or choose by text:

    ![image](https://cloud.githubusercontent.com/assets/104830/10479273/ac84a36e-7263-11e5-9e25-db09c8529510.png)

2. Or preview the topology from the select box:

    ![image](https://cloud.githubusercontent.com/assets/104830/10479385/68fc0974-7264-11e5-833e-d8d763d73e4d.png)

More ideas? What do you think @ChaelKruip @antw @grdw ?

grdw:
Maybe a dropdown with images:

![dropdown-with-images](https://cloud.githubusercontent.com/assets/2676542/10479993/eb1d839e-7267-11e5-9d9d-83df6a583906.png)

dennisschoenmakers:
> Maybe a dropdown with images:

Good idea! Would the images be large enough to be understood though?

grdw:
> Good idea! Would the images be large enough to be understood though?

It's only to get a basic idea of the 'shape' of the topology. Which for Lochem topologies might become a bit of a problem when it comes to size.

I personally like the "preview idea" a bit better than my own when I think a bit harder about the usability problem. Would also make the topology preview - not just a preview - but a bit more interactive. For instance: seeing which stakeholders are attached and maybe some advanced properties when you hover/click the node.

antw:
Nice idea. :+1:

> Would also make the topology preview [...] seeing which stakeholders are attached and maybe some advanced properties when you hover/click the node.

Funnily enough, my first thought was in the opposite direction: strip out as much information from the topology preview as possible: reduce the nodes to a 3-4px wide, hide the labels, and bring the nodes closer together (kind of like in [the drop-down version](https://github.com/quintel/etmoses/issues/530#issuecomment-147990821)). That way we'd fit the diagram into a smaller space, and might stand a better chance of displaying the very large networks.

It's pretty easy to change attributes of the topology later (capacity, stakeholder, costs, etc); I don't think it's important that those are visible in the preview. Those could be shown in-detail on `TopologiesController#show`.

JeroenvdLogt:
[...]

PR #539, opened 2015-10-19
Topology preview and other fixes

grdw:
This pull request contains the solution for #530, #537


antw:
* I removed the global `depths` variable which wasn't necessary. If you were to call `depthCount` without first setting `depths` to zero, you'd get an incorrect value back. [Globals are almost always evil](http://programmers.stackexchange.com/a/148109). I can explain my changes in more detail if that would be helpful?

* It would be nice if the JS files would pass [JSLint](http://www.jslint.com/), or have a valid reason for *not* passing (some of the rules it applies are over-kill, but many are useful), such as by using spaces before brackets `function thing(){}` → `function thing() {}` and adding missing semi-colons. The same suggestion about brackets is true of Ruby also (`lambda{}` → `lambda {}`), but that doesn't apply to this PR.

* Would be preferable to have one "thing" per pull request, and 567c59b94ac8320c5da7c6c1b0dd5fd7e71ee15d has two unrelated changes together ("economic_lifetime" and CSS changes).

The topology preview is *really* neat; I think people will love it. :+1:

app/assets/javascripts/topology_preview.js
```diff
@@ -1,33 +1,34 @@
...
   TopologyPreviewer.prototype = {
     preview: function(){
       clear();
 
-      svgHeight = viewerHeight();
-      svgWidth  = viewerWidth();
-      svg       = buildBaseSvg();
-      data      = [graphData()];
-      root      = data[0];
+      currentStylesheet = TopologyStylesheet[style];
```
antw: No `var ...`? Are all globals?

grdw: They're defined as a `var` on line 2. Than filled at line 18 until 22. 
They're not globals. See it as this:

````javascript
var A = (function(){
  var b;

  b = 5;
})();

console.log(b);
````

It's a global within the scope of `A` which makes it local. It might also be that I'm confusing your definition of global with my definition of global: http://www.w3schools.com/js/js_scope.asp

Comment (antw) [inline app/assets/javascripts/topology_preview.js]: What I mean is, if I do:

```js
new TopologyPreviewer(...)
new TopologyPreviewer(...)
```

... each instance doesn't get it's own variables; won't the vars used by the second Previewer overwrite the vars of the first? Putting them inside a closure means they're not globally *readable*, but still only one can exist at any moment in time, right? Maybe I'm misunderstanding?

antw: It reminds me of [this style](https://gist.github.com/antw/3c86b4df34281d768ca5) of creating JS classes. Perhaps that's what you were going for?

grdw: > overwrite the vars of the first? Putting them inside a closure means they're not globally readable, but still only one can exist at any moment in time, right? Maybe I'm misunderstanding?

Yes. That's what happens. This is mainly due to the fact that there's only one previewer actually doing the `previewing`. But I'll change it to the `this.`-notation. To give each new object their own set of variables.

I'm using a bit of a mash-up from two design patterns which was once advised to me in a fancy blog article which I can't find anymore.

```javascript
var Car = (function(){ 
  Car.prototype = { 
     public: method(){ 
       return 5;
     } 
  };
   
  function private(){
  };
   
  function Car(){
  };
  
  return Car;
})();
````

Also, I think `Coffeescript` 'converts' it in such a fashion.

Comment (antw) [inline app/assets/javascripts/topology_preview.js]: > Also, I think Coffeescript 'converts' it in such a fashion

Indeed, CS gives:

```js
// class TopologyPreviewer
//   constructor: (@topologyGraph, @presetData, @style) ->

var TopologyPreviewer;

TopologyPreviewer = (function() {
  function TopologyPreviewer(topologyGraph, presetData, style) {
    this.topologyGraph = topologyGraph;
    this.presetData = presetData;
    this.style = style;
  }

  return TopologyPreviewer;

})();
```

The key is that all the (instance) variables are assigned in the `TopologyPreviewer` constructor.

> but I'm not a big fan of Coffescript

Perhaps we should look into using ES6 with Babel for future projects? ES6 would solve all these problems for us. :grinning:

grdw: > Perhaps we should look into using ES6 with Babel for future projects? 

Yes! :+1:  Would be nice. Never heared of Babel before though. **Is now looking at babeljs.io**

grdw: But back on topic; could this be merged :angel: ?

antw:
> could this be merged

:+1:

PR #548, opened 2015-10-20
Adding 24 customer variations

grdw:
- Updated the business case summarizer
- Updated the comparison tool
- Abstracted hierarchy to be implemented into stakeholders aswell.


antw:
The "incoming" amounts don't seem to be showing up in the business case table. I have a market model which looks like:

* customer → supplier
* system operator → government

... but the table looks like:

screen shot 2015-10-21 at 16 40 11

The "supplier" and "government" stakeholders are missing, even though they should be receiving money. This only happens when the `stakeholder_to` doesn't have any outgoings of their own. If I also add:

* system operator → customer

... the incoming amount from that interaction *is* shown:

screen shot 2015-10-21 at 16 45 29

grdw:
> This only happens when the stakeholder_to doesn't have any outgoings of their own

I have this market model:

screen shot 2015-10-22 at 9 38 32 am

And this business case:

screen shot 2015-10-22 at 9 40 27 am

If I were to follow my train of thought:

- The `in`-column for `customer` should be filled but it isn't. 
- And the reason for that is, that "the customer doesn't have any outgoings".
- And that's stupid because a customer does receive money from... ?

grdw:
I did found out though, that the stakeholders list wasn't right when calculating the financials.

I'm still not really sure if that was the thing that killed it.

antw:
> I'm still not really sure if that was the thing that killed it.

That does appear to have fixed it, thanks.

--

> And that's stupid because a customer does receive money from... ?

In your example, the customer would not have any incoming money. Here's the full market model and business case which *was* broken (now fixed):

**Market model**

screen shot 2015-10-22 at 09 57 41

**Business case**

screen shot 2015-10-22 at 09 57 53

The supplier and government were both missing, despite receiving money from the customer and system operator respectively.

dennisschoenmakers:
I can't seem to figure out how to use the different (24 different) stakeholders. Can we add something to the README? cc @ChaelKruip

Issue #555, opened 2015-10-21
Importing gas technologies


grdw:
Will fill out this issue when more is clear.

ChaelKruip:
> when more is clear.

Let me clarify: just as with PV, EV and four types of heat pumps, we need to 'import' the number of some gas-fuelled space heating and hot water technologies from the ETM into ETMoses. The following technologies are relevant:

* households_space_heater_network_gas
* households_space_heater_combined_network_gas
* households_space_heater_micro_chp_network_gas (import but don't show to user yet because this needs some extra consideration, see https://github.com/quintel/etmoses/issues/556)
* households_water_heater_network_gas
* households_water_heater_combined_network_gas
* households_water_heater_micro_chp_network_gas (import but don't show to user yet because this needs some extra consideration, see https://github.com/quintel/etmoses/issues/556)
* households_water_heater_fuel_cell_chp_network_gas (import but don't show to user yet because this needs some extra consideration, see https://github.com/quintel/etmoses/issues/556)

grdw:
> The following technologies are relevant:

Thanks for the technologies. :smile:  I put them in a migration and they're in the database. And set visible to false for the three technologies you just specified. 

I'm guessing that this is not the only thing that needs to happen? :sweat_smile:

ChaelKruip:
> I'm guessing that this is not the only thing that needs to happen?

They need to show up in the technology matrix as well... Later, they need to be coupled to the heat demand but that is not part of this ticket :smile:

grdw:
> They need to show up in the technology matrix as well

That happens automatically with all the visible technologies :+1:

ChaelKruip:
Can we close this @grdw ?

grdw:
Yes

PR #560, opened 2015-10-22
Adding gas technologies

grdw:
Nothing to fancy.

Fixes: #555


antw:
@ChaelKruip Is this to be merged now? I'm guessing the technologies aren't of much use yet if we have no profiles or behaviours for them?

ChaelKruip:
> @ChaelKruip Is this to be merged now? I'm guessing the technologies aren't of much use yet if we have no profiles or behaviours for them?

I agree. Let's keep this PR open for now. Just good to know that the actual work is done.

PR #571, opened 2015-10-26
Make it possible to download the merit order

dennisschoenmakers:


antw:
1. Did you mean to add `spec/fixtures/points.tmp.csv`? It doesn't seem to be referenced anywhere in the tests.

2. I'm not sure about the merit Rails action returning the curve as JSON, and then having the JavaScript rewrite that as a CSV. That seems a bit unnecessary as the front-end likely won't need the raw merit price data. Is there a reason why the Rails action can't just [send the CSV](http://nithinbekal.com/posts/rails-csv-export/) to be downloaded by the browser?

  *(That's not done for the node load CSVs as 1. the server can't easily get the load of an individual node without recalculating the network (will change with caching); and, 2. the front-end already has the loads of each node already available).*

grdw:
> Did you mean to add spec/fixtures/points.tmp.csv? It doesn't seem to be referenced anywhere in the tests.

No. It's removed now.

> I'm not sure about the merit Rails action returning the curve as JSON, and then having the JavaScript rewrite that as a CSV. That seems a bit unnecessary as the front-end likely won't need the raw merit price data. Is there a reason why the Rails action can't just send the CSV to be downloaded by the browser?

Well the reason was the load calculation. Did manage to get around that though with a 'hidden' form.

antw:
> Well the reason was the load calculation. Did manage to get around that though with a 'hidden' form.

Ah yes, makes sense. That might be worth documenting in the controller action.

With caching now merged, I think this might not be necessary anymore; we can read the load of the top-most node out of the cache. If that'll take too long, I suggest we merge this and come back to using the cache data later. What do you think?

grdw:
> I suggest we merge this and come back to using the cache data later. What do you think?

Yes. I changed your 'No-no' comment. I put the merit actions in a separate contoller (also to avoid making the `testing_grounds_controller` one pile of actions). If you could take a look.

antw:
Works nicely in some cases, but other testing grounds throw this error at me:

```
JSON::ParserError - 399: unexpected token at '':
  json (1.8.2) lib/json/common.rb:155:in `parse'
  app/controllers/merit_controller.rb:22:in `create_merit_order'
  [...]
```

The two scenarios where I've seen this so far have been local copies of:

* http://beta.moses.et-model.com/testing_grounds/351
* http://beta.moses.et-model.com/testing_grounds/349

grdw:
> The two scenarios where I've seen this so far have been local copies of:

I mported the staging database into my local development database. But I can't reproducte the same result. Could you show me your development.log params? 

The thing I did saw was that the `altLoad` was not used when available.

antw:
[Uploaded the relevant parts of the log here](https://gist.github.com/antw/1d506e0f7a78f568465c).

It appears the `base_load` param is truncated for some reason. If I add the missing `]` and parse the result, there are 28339 elements in the resulting array instead of 35,040.

In case it's relevant, I'm using Chrome 46.0.2490.80.

grdw:
> In case it's relevant, I'm using Chrome 46.0.2490.80.

Yes, it's relevant :smile: 

I can only trigger the error in Chrome 46.0.2490.80. I'm using Firefox 41.0.2. 

Maybe Chrome has a (lower) limit for the size of the POST body.

grdw:
Allright, take three. :clapper: 

Right now I grab the root load from caching instead of the big request.

antw:
Using the cached data is good, but I think the way it's done is not quite right.

* Using NetworkCache::Fetcher in the MeritCurveBuilder breaks the idea of "transparency".

  I'm definitely regretting using the work "transparency" since it doesn't really explain the idea well. Maybe I can put it more simply as: the words `NetworkCache` should never appear in the front-end (JavaScript), or in the `Network` or `Market` namespaces. If it does, those *separate modules* become dependant on the existence of a cache.

* Reading directly from the cache may fail if the cache data hasn't been written. This is unlikely to happen in everyday use, but not impossible.

It should probably look something like:

```ruby
class MeritController
  private

  def merit_order
    network = TestingGround::Calculator.new(@testing_ground, params[:strategies]).calculate
    Merit::MeritCurveBuilder.new(@testing_ground.scenario_id, network[:graph]).merit
  end
end
```

grdw:
> It should probably look something like:

:clapper:  4! :smile:

Issue #385, opened 2015-09-14
Some LES's take incredibly long to load causing a 504. 


grdw:
![screen shot 2015-09-14 at 3 54 58 pm](https://cloud.githubusercontent.com/assets/2676542/9851453/020db898-5af9-11e5-9f2e-9d12ccdcb788.png)

antw:
The default Nginx `proxy_pass` timeout is 60s, which would explain the 504. We could increase that setting, but maybe the better solution is to move the calculation into a background worker once #377 is merged.

grdw:
I do have a question about this though. Where would I store the load calculation when the background processor is finished with it's calculation? With a business case we just use the financials table. With this calculation I'm not to sure we should use the database for this.
`Rails.cache`?

What are your thought, @antw ?

grdw:
This is currently in a branch with the same issue number. Waiting for pull request #377 before making a merge request.

antw:
> Where would we store the load calculation **when the background processor is finished** with it's calculation? With this calculation I'm not to sure we should use the database.

For this reason alone, I'd suggest not proceeding too far with this, for now (5MB for a *small* LES). I think there is a good chance that there is simply too much data to put into a "dumb" cache and hope for the best. As we will also have to start storing individual technology loads and storage amounts, the requirements may increase by 10-100 times.

Caching is something we need to think about properly; we'll have to test and benchmark the off-the-shelf options like Memcache, Redis, etc, but may well end up with a home-made serialize-to-disk solution ([ETEngine does this](https://github.com/quintel/etengine/blob/master/app/models/etsource/atlas_loader.rb#L90-L111), though it's requirements are considerably simpler than Moses' are).

Sticking it in Rails.cache won't work: "staging" *currently* uses an in-process cache limited to 64MB. The calculated LES will end up in the delayed_job process cache, not the Rails one.

grdw:
> Sticking it in Rails.cache won't work

Okay. Thanks for clarifying.

>Caching is something we need to think about properly; we'll have to test and benchmark the off-the-shelf options like Memcache, Redis, etc, but may well end up with a home-made serialize-to-disk solution

Yes.

grdw:
@antw I'm just thinking out loud here and this may sound a little far off.

But couldn't we - instead of requesting the whole load calculation at once - request it in 'parts'. 
[...]

PR #575, opened 2015-10-28
Issues/#519 calculate les bg

grdw:
Calculate the load in delayed job.

Fixes: #385

### DIFF
--- a/app/assets/javascripts/ajax.js
+++ b/app/assets/javascripts/ajax.js
@@ -1,0 +1,17 @@
+var Ajax = (function () {
+    'use strict';
+
+    return {
+        json: function (url, data, success, error) {
+            $.ajax({
+                type:        "POST",
+                contentType: "application/json",
+                dataType:    "json",
+                url:         url,
+                data:       JSON.stringify(data),
+                success:    (success || function () { return; }),
+                error:      (error || function () { return; })
+            });
+        }
+    };
+}());
--- a/app/assets/javascripts/business_case_validator.js
+++ b/app/assets/javascripts/business_case_validator.js
@@ -1,26 +1,32 @@
-$(document).on("page:change", function(){
-  var newImportForm = $("form#new_import");
+var BusinessCaseValidator = function (form) {
+    'use strict';
 
-  if(newImportForm.length > 0){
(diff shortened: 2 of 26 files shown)


antw:
Is this ready? There are a lot of (I think) unrelated changes, so it's hard to tell. I *think* it might need to be updated or rebased on account of the changes that were made to the caching PR; I see [some code from the original implementation](https://github.com/quintel/etmoses/commit/c472317f33ca43ed20794ba95b9f5330474978f1#diff-27a7aedba59d5f302ac9ff620caf0273R22) in this one.

grdw:
> Is this ready?

Yes. The thing is that I originally created the branch of master when I was done with the first version of the caching. Than when the changes occured I had to do a really painful rebase with lots of conflicts. 

Now the rebasing doesn't do that much. Maybe I could remove that one commit because it doesn't add that much value for this PR.

antw:
I tried this locally, but it's not working for me: it seems to expect a `job_id` column in `testing_grounds`, but there's no migration to create it. I'm curious if a single `job_id` column is compatible with the idea of calculating the "basic" and "feature" networks simultaneously (in separate jobs)?

grdw:
> but there's no migration to create it. 

Crap, think I deleted that migration with the commit I removed from this PR.

grdw:
> I'm curious if a single job_id column is compatible with the idea of calculating the "basic" and "feature" networks simultaneously

Well, yes. I was curious that it *did* work aswell. My guess is though that my local Rails environment `queues` up the two requests. 

I added the migration :+1: so you can fiddle with it.

antw:
> My guess is though that my local Rails environment queues up the two requests.

I think so too. It might not work in a multi-process environment such as we have on staging and production. Separate `basic_job_id` and `feature_job_id` columns would do the trick, as would storing these in a separate table with `testing_ground_id`, `job_id` and `type` columns (where `type` is "basic" or "feature").

A separate table is more flexible should we ever add more "types" (for example, a low resolution version), but I doubt that will happen in the *immediate* future.

grdw:
> as would storing these in a separate table

Did it. :smile:

PR #617, opened 2015-11-20
Issues/#604 high res image of topology

grdw:
Works in IE9! 

Hurray!

:dancers:


app/controllers/topologies_controller.rb
```diff
@@ -59,6 +60,19 @@ def destroy
     redirect_to(topologies_url)
   end
 
+  # POST /topologies/:id/download_as_png
+  def download_as_png
+    filename = "#{ Rails.root }/tmp/#{ @topology.id }.png"
```
antw: I would prefer these go in a subdirectory, instead of being dumped into `tmp`. I suggest something like `tmp/topology_images`. Using [Tempfile](http://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html) is often cleaner, since you don't have to worry about where you save the image.

I think even better (haven't tested this myself, but it should work) is to have Rmagick generate the image and send it to the browser without having to save it to disk first:

```ruby
# ...
img[0].resize!(2)

send_data(
  img[0].to_blob(format: 'PNG'), 
  content_type: 'image/png', disposition: 'inline'
)
```

antw:
I'm not having much luck with this in Chrome or Firefox. The PNG contains content like:

```
push graphic-context
viewbox 0 0 1140 300
affine 1 0 0 1 0 0
push graphic-context
affine 1 0 0 1 100 15
# ...
```

grdw:
> I think even better (haven't tested this myself, but it should work) is to have Rmagick generate the image and send it to the browser without having to save it to disk first:

Yes :smile:  Thanks, that's awesome. Do have an issue with downloading the image though.. 
I can copy and paste it to let's say GIMP. But I can't download it from the browser.

grdw:
>push graphic-context
viewbox 0 0 1140 300
affine 1 0 0 1 0 0
push graphic-context
affine 1 0 0 1 100 15

How would I manage to get this result?

In my case the image is downloaded and I can see it. And that's it.

antw:
> How would I manage to get this result?
> In my case the image is downloaded and I can see it. And that's it.

Never mind, it's working now! It might have been a change you made in the most recent commit, or that I reinstalled ImageMagick; either way it works now. The image is quite bit though, is that intended?

(2280 x 800):
download_as_png

Last suggestion: Include a `filename` when using `send_data` to give the PNG a more meaningful name (it defaults to "download_as_png.png"):

```ruby
send_data(
  img[0].to_blob { self.format = 'PNG' },
  filename: "#{ topology.name }.png",
  content_type: 'image/png'
)
```

grdw:
>  The image is quite big though, is that intended?

Yes. The client wants to include it in presentations. So a 'larger' topology would be nice.

> Include a filename when using send_data to give the PNG

Will do.

antw:
> The client wants to include it in presentations. So a 'larger' topology would be nice.

I see, it's a result of `resize(2)`. I don't see the point, personally. It just increases the size 2x, and the result is a big, blurry image. The client can accomplish the same thing by scaling the image up or down in their presentation software.

If a way can be found to rescale the image without losing detail or becoming blurry, that would be pretty nice; otherwise I don't think it's worth it.

grdw:
> If a way can be found to rescale the image without losing detail or becoming blurry, that would be pretty nice; otherwise I don't think it's worth it.

Without the `resize(2)` the image still looks pretty blurry. I did remove it though and increased the [quality](https://rmagick.github.io/info.html#quality) to 100.

PR #618, opened 2015-11-23
Removes all NaN's for conditional consumption

grdw:
@antw  please have a look because this looks rather patchy. 

It's for the error generated here: http://beta.moses.et-model.com/testing_grounds/143

It does fix the problem but I must say that I have no clue how it get's those `NaN`'s in the first place. Maybe you know a better solution than this patch fix?

### DIFF
--- a/app/models/network/path.rb
+++ b/app/models/network/path.rb
@@ -55,7 +55,7 @@ module Network
     #
     # Returns a numeric.
     def available_capacity_at(frame)
-      @path.map { |node| node.available_capacity_at(frame) }.min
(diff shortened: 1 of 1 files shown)


antw:
Thanks. I won't merge this since discarding `NaN`s doesn't really solve the problem; it only covers it up. We used to silently discard exceptions in ETEngine and as a result it became a nightmare to debug (this was remedied some time ago), so I don't want to do it here.

Either the error is in my `Network` code or in the data given by the user. If it's the former, I need to find and fix it; if the latter it indicates we need to add a validation to prevent reoccurrences.

I've assigned #608 to myself and I'll investigate when I can.

PR #621, opened 2015-11-24
Setting new attributes for CSV upload

grdw:


antw:
Looks good, but I get `ActionController::InvalidAuthenticityToken` when importing the technology CSV. Also happens on beta, so it isn't caused by a change in this pull request.

Issue #629, opened 2015-11-27
Uploading technology csv results in name: other


JeroenvdLogt:
When I try to upload a technology csvfile this results in a table with all the name cells as other. I will snet the csv file to Chael by e-mail. The topology I use is USEF: waardebepaling flex (Lochem doorrekening) at the beta server.

JeroenvdLogt:
Another thing I encounter, when I change name: other in name: electric car f.e. all the previous settings of capacity-storage-annual usage are set to default automatically. It cost a lot of time to reset everything to the previous values. Is it possible to cancel default values?

dennisschoenmakers:
Due to the holiday of @ChaelKruip assigned this to @grdw

PR #633, opened 2015-11-27
Fixing type column for csv

grdw:
Fixes: #629

### DIFF
--- a/app/models/technology_list.rb
+++ b/app/models/technology_list.rb
@@ -62,6 +62,12 @@ class TechnologyList
     JSON.dump(Hash[list.to_h.map { |key, techs| [key, techs.map(&:to_h)] }])
   end
 
+  def self.technologies_in_csv(parsed)
+    names = parsed.map{|row| row['name'] }.uniq
+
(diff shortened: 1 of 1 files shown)


antw:
Same problem as #621; I get a `ctionController::InvalidAuthenticityToken` when I choose a CSV to upload and hit "save". Is it just me? :anguished:

grdw:
> when I choose a CSV to upload and hit "save". Is it just me?

I'm having a hard time reproducing this one. Because importing a CSV file and than calling create calls to `TestingGroundsController#create as HTML` with an actual valid authenticity token as far as I know.

antw:
> calling create calls to `create` as HTML with an actual valid authenticity token as far as I know.

Exactly! It's weird; maybe it's something in my local setup. If you know that this works, let's merge it. :+1:

grdw:
Do you use pow? https://github.com/basecamp/pow/issues/96 

Sort of did a really quick google.

antw:
> Do you use pow?

Yeah; restarting Pow has no effect, and the exception also gets raised when using the built-in Rails server. It's got to be something local, never mind.

Issue #632, opened 2015-11-27
few seconds waiting for dropdownmenu in technology tale


JeroenvdLogt:
If i click on a dropdownmenu of f.e. name in the technology table it takes ET-Moses a few seconds (2-4) to show the dropdownlist.

grdw:
Could you copy/paste a link to a LES where this happens?

dennisschoenmakers:
Could this be particular to IE?

Sent from my iPhone

On 27 nov. 2015, at 13:58, Gerard Westerhof <notifications@github.com>
wrote:

Could you copy/paste a link to a LES where this happens?

—
Reply to this email directly or view it on GitHub
<https://github.com/quintel/etmoses/issues/632#issuecomment-160134261>.

JeroenvdLogt:
i am using google chrome

dennisschoenmakers:
> i am using google chrome

Wise decision! :smile: Thanks for telling, then we know it cannot be a cross browser thing.

JeroenvdLogt:
http://beta.moses.et-model.com/testing_grounds/376/edit

hereby a link of a les where I encounter the problem

dennisschoenmakers:
Here it works very snappy on Chrome... @grdw can you reproduce?

On Fri, Nov 27, 2015 at 8:35 PM, Jeroen van de Logt <
notifications@github.com> wrote:

> http://beta.moses.et-model.com/testing_grounds/376/edit
>
> hereby a link of a les where I encounter the problem
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/quintel/etmoses/issues/632#issuecomment-160194267>.
>



-- 

Dennis Schoenmakers, *Partner*


*Quintel Intelligence*Atrium - Strawinskylaan 3051
1077 ZX  Amsterdam
The Netherlands

*Visiting Address:*
Keizersgracht 639
1017 DT Amsterdam

T  : +31 (0)20 301 22 69
F  : +31 (0)20 301 22 02
M : +31 (0)6 109 65 293

E : dennis.schoenmakers@quintel.com
I  : www.quintel.com and www.energytransitionmodel.com

*This message (including any attachments) is confidential and may be
privileged. Any review, retransmission, dissemination or other use of, or
taking any action with reference to this information by persons other than
the intended recipient is strictly prohibited. If you have received it by
mistake please notify the sender by return e-mail and delete this message
from your system. Please note that e-mails are susceptible to change. The
[...]

PR #634, opened 2015-11-30
Rewriting bits of editable table to improve speed

grdw:
This is to patch #632 to improve the speed by a little.

I'm not going to put too much time into this since we're going to the new interface anyway. You can also decide to not merge this because of that. Either way I'm fine with it.

### DIFF
--- a/app/assets/javascripts/tables/editable_table.js
+++ b/app/assets/javascripts/tables/editable_table.js
@@ -18,15 +18,17 @@ var EditableTable = (function () {
     }
 
     function tableHeader(index) {
-        return $($(this.selector).find("thead th")[index]).data("header");
+        return $(this.selector).find("thead th").get(index).getAttribute("data-header");
     }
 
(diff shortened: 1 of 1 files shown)


antw:
I'm not really seeing any improvement in Chrome; is it better in Firefox? I hope to merge the new UI today so I wouldn't worry about this too much, unless the problem is also present there.

grdw:
>  is it better in Firefox? 

A little bit, it scrapes of a second which is not really noticable in the scheme of things. My other guess would be that it has to do with the overall finding of all the input elements and building up the whole `technology_distribution` object. 

The new interface doesn't look for all the individual input elements. Instead it ust loops over all `.technology`-elements and grabs all the data attributes, which saves a lot of DOM searching. The input elements - when changed - write to the data attribute of their `.technology`-element parent.

In the new interface I don't have any 'lag'. It's a little bit slow with large LES'es but not as slow as the current editor.

PR #646, opened 2015-12-07
New technology matrix - Part 2

grdw:
This is not yet 'stable'.

There are some known problems with this.


antw:
Thanks man! Is this *done* or still a work-in-progress? If it is finished, I'll rebase the gas-networks branch against it and merge them at the same time.

grdw:
> Thanks man! Is this done or still a work-in-progress?

Well there's still some bugs in the pipeline (see last three issues). :cold_sweat:  Some of them I find to be pretty hard.

antw:
No worries. The gas-network branch – with the basic gas techs, and boosting/buffering – isn't done yet (end of today probably) so there's no rush. I think I'll wait a bit before rebasing otherwise it'll probably have to be done again once there are other changes.

> :cold_sweat: Some of them I find to be pretty hard.

I would say, don't worry too much if the code gets a bit "hacky" in places; we can clean it up later. :smiley:

grdw:
@antw YES!

I fixed the most annoying bugs. There's maybe some thing still not right for the demand of the heat pumps that are attached to the buffers because I haven't found the time to actually look into much detail of this.

antw:
Wonderful! I'll get this merged into the gas-networks branch, and then on to master, ASAP. Thanks!

antw:
@grdw I found a couple of minor issues. All of these are when *editing an existing LES*, I haven't yet tried importing a new one.

FYI: Your work has been rebased onto the `gas-networks` branch; if you are able to fix any of these, please either do it on that branch, or on a *new* branch based on it. Basically, treat `gas-networks` as if it is (temporarily) `master`.

* When I add a new space heating or hot water buffer, it appears that the `profile` and `profile_key` attributes are not saved unless I first click on the "Profile" drop-down and change it's value before pressing "Save".

* I think if I add a heat pump to an endpoint which does not already have an appropriate buffer, then a buffer should be added, and the heat pump should belong to it. I'm open to alternate ideas, but I don't think will make sense in future LES' to have stand-alone heating and hot water technologies.

  It is okay for *old* LES' to keep the stand-alone technologies they already have, as the network calculation is backwards-compatible with them, but *new* techs (the hybrid heat-pump) won't work unless attached to a buffer.

  @ChaelKruip What do you think?

* The first buffer I add is called "Buffer space heating NaN". If I add more they are numbered correctly: "Buffer space heating 2", "Buffer space heating 3", etc. Not a big deal.

ChaelKruip:
> @ChaelKruip What do you think?

I agree!

grdw:
> When I add a new space heating or hot water buffer, it appears that the profile and profile_key attributes are not saved unless I first click on the "Profile" drop-down and change it's value before pressing "Save".
> The first buffer I add is called "Buffer space heating NaN". If I add more they are numbered correctly: "Buffer space heating 2", "Buffer space heating 3", etc. Not a big deal.

Are fixed.

We are going to disable the 'single heat pumps' if there are no buffers present for them. I'm currently fixing that.

grdw:
> but new techs (the hybrid heat-pump) won't work unless attached to a buffer.

We solved it as described in the previous issue. :smile:

antw:
This has been merged (manually) into master; thanks!

PR #662, opened 2015-12-10
Calculating initial costs with units count

grdw:


antw:
:+1:

dennisschoenmakers:
Thanks! Could you add some documentation to the [**topology_costs**](https://github.com/quintel/etmoses/blob/master/app/models/market/initial_costs.rb#L24-L29) and [**technology_costs**](https://github.com/quintel/etmoses/blob/64f973e221cb34437b4bc203103fc5fbbf36d4f8/app/models/market/initial_costs.rb#L31-L37) methods? explaining the difference?

Issue #661, opened 2015-12-09
Issues with the new technology editor


antw:
I have merged the new technology editor and gas network into master (*not* deployed to staging), but there are some minor "niggles" still occurring. I'll note them here, but I can split this into separate issues if preferred.

* **I can't add certain technologies to an existing LES.** I can *delete* and edit technologies successfully. Strangely, some technologies *can* be added: 

  Can be added:
  * Battery
  * Power-to-gas

  Cannot be added:
  * Buffer space-heating
  * Buffer hot water
  * Household
  * Building
  * Electric vehicle
  * Other
  * Residential PV panel 

  It seems that I *can* import a new LES with these technologies, but I cannot add them manually to an existing LES.

* **Adding a household technology shows two profile select boxes:**

  ![screen shot 2015-12-09 at 18 06 21](https://cloud.githubusercontent.com/assets/4383/11695009/edbeb96a-9ea4-11e5-80de-c66b3129ac0f.png)

* **Using the default topology, if I add a battery or power-to-gas to the "Households 2" endpoint, the technology is instead added to "Households 1"**. This only seems to affect existing testing grounds; technologies are assigned to Households 2 and 3 *and successfully saved* when importing a *new* LES.

grdw:
>  Adding a household technology shows two profile select boxes:

You should change the `units` field for some duplication fun:

![screen shot 2015-12-10 at 8 57 10 am](https://cloud.githubusercontent.com/assets/2676542/11709959/05b0f124-9f1c-11e5-8c31-766c98b0dcbe.png)

:sweat_smile: 

I'll look into the issues though.

grdw:
> I can't add certain technologies to an existing LES. 

"but I cannot add them manually to an existing LES."

Do you have an exmample LES?
Or what do you mean by 'added'. Does that mean, that the template does pop-up but it doesn't add it to the `technology_profile` in the database?

> Adding a household technology shows two profile select boxes: -> fixed

> Using the default topology, if I add a battery or power-to-gas to the "Households 2" endpoint, the technology is instead added to "Households 1"

I still need to look into this one.

antw:
> You should change the units field for some duplication fun:

:laughing:

> Does that mean, that the template does pop-up but it doesn't add it to the `technology_profile` in the database?

Yes exactly. The UI works fine, the data is sent to the Rails app correctly, but the TestingGround doesn't get saved.

[...]

Issue #642, opened 2015-12-01
Name does not show up correctly in warning


ChaelKruip:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/11500262/4d50c37a-982c-11e5-9cd4-18c5060c3456.png)

grdw:
It's because most users don't have a name... They should be able to edit that though or a simpler fix would be to display an email address and remove the name column.

ChaelKruip:
> a simpler fix would be to display an email address and remove the name column

Or display the email address only if there is no name

antw:
There are privacy implications to displaying an e-mail address publicly, and [our privacy policy](http://et-model.com/privacy) is quite explicit that they're only used to associate users with their scenarios.

I would suggest adopting the name "Anonymous" if a user does not have a name, or perhaps display a different version of the warning message: "You are now editing someone else's LES".

ChaelKruip:
> There are privacy implications to displaying an e-mail address publicly, and our privacy policy is quite explicit that they're only used to associate users with their scenarios.

Good point!

> I would suggest adopting the name "Anonymous" if a user does not have a name

I like!

grdw:
> display a different version of the warning message: "You are now editing someone else's LES".

This has my preference :+1:

PR #665, opened 2015-12-10
Issues/#627 the buffer and the booster

grdw:
Fixes: #661, #642


antw:
Works perfectly, thanks @grdw!

PR #682, opened 2015-12-11
Issues/hybrid heat pump and misc.

grdw:
This is a bit of a mash-up PR. I know I shouldn't do it; for future reference.

- But this contains migrations for the hybrid heat pump and a minor class in the import folder to split that hybrid up in multiple technologies.
- Also this contains the renaming of `capacity`.
- And `temperature_profiles`.


app/assets/javascripts/les/profile_select_box.js
```diff
@@ -36,7 +36,7 @@ var ProfileSelectBox = (function () {
             key = defaultValue.replace(/default/, '').toLowerCase();
 
             if (key === 'capacity') {
-                key = 'electrical_capacity';
+                key = 'carrier_capacity';
```
grdw: @antw would these three lines still be needed now that's decided that "Output capacity" (`capacity`) is displayed in the technology matrix instead of `electrical_capacity`?

antw: I don't think so. It should be safe to remove them.

antw:
Replaced by #689, which has been merged. Thanks!

PR #694, opened 2015-12-15
Fixing bug with business calculator

grdw:


antw:
I'm not sure how the version I wrote ever worked. :confused:

This isn't quite ideal though. The "featureless" networks are only needed to calculate the "flexibility" measures; the aim of the original version was to only calculate the featureless networks when they're actually used. If there are no flexibility measures, the extra calculation never gets done.

This PR changes it so that the featureless network is *always* calculated.

Here's what I'd suggest:

* Change Market::Variant so that the `object` method is public.
* Change the BusinessCaseCalculator to have something like:

   ```ruby
    # Lazily computes the featureless networks only when required.
    featureless = Market::Variant.new do
      @testing_ground.to_calculated_graphs
    end

    electricity = Market::Variant.new do
      featureless.object.detect { |net| net.carrier == :electricity }
    end

    gas = Market::Variant.new do
      featureless.object.detect { |net| net.carrier == :gas }
    end
   ```

grdw:
> This PR changes it so that the featureless network is always calculated.

Yes. That was not supersmart...

I fixed it as described in your suggestion.

antw:
Perfect, thanks!

PR #699, opened 2015-12-16
Make it possible to change your own password + name

grdw:


antw:
I'm a bit confused why we aren't just using the Devise controllers/templates for this?

The user model has validations for e-mail address and password, also courtesy of Devise, so the `UsersForm` (and existing `RegistrationForm`) seem redundant to me.

grdw:
> I'm a bit confused why we aren't just using the Devise controllers/templates for this?

Well, I think it has to with possible future 'custom' features and general distaste for devise due to the 'magic'. Ofcourse this is quiet farfetched but from experience there was usually a case where a customer wanted something custom concerning signing in and signing up which devise didn't support. 

Ofcourse, for the state 'the signing in / signing up' situation is currently, devise would be fine.

PR #702, opened 2015-12-17
Using a volume in liters for buffers instead of the default volume

grdw:
I'm going to review this first.


grdw:
@ChaelKruip how desperate do we need this? I'm asking because this adds two new columns with numbers to the *database* which is something we would like to see changed in the nearby future (I hope).

ChaelKruip:
> @ChaelKruip how desperate do we need this? I'm asking because this adds two new columns with numbers to the database which is something we would like to see changed in the nearby future (I hope).

Not very urgent.

PR #710, opened 2015-12-22
Adding the neighbourhood battery to the front end

grdw:


antw:
@grdw I have rebased this against master to account for your other recent changes there, **and force-pushed to the pull request branch**.

I did notice a couple of issues:

1. If I add a neighbourhood battery to an LES, save, and then refresh the edit page, the wrong template is used (the profile `<select>` is visible and the slider is not rendered):

  screen shot 2016-01-08 at 17 04 22

  (It is also highlighted in red indicating an error, as it has no profile. This also affects the normal battery *so this does not need to be fixed with this pull request*).

2. If I visit the edit page and add a neighbourhood battery, the slider tooltip is not aligned with the slider handle:
  
  screen shot 2016-01-08 at 17 06 23

  It jumps to the correct position after moving the handle.

3. I would prefer for the slider value to be always visible; for example:

  screen-shot-2016-01-08-at-17 11 19

(2) and (3) are not particularly important; we can merge this without them. (1) is a bit of a problem.

grdw:
> 2. If I visit the edit page and add a neighbourhood battery, the slider tooltip is not aligned with the slider handle:

I know about this one. I think this is a bug with the bootstrap slider. I can't find any CSS rule that's confusing the slider-head. The same problem also exists for the solar capping slider. If you drag it just a little it automatically aligns properly.

> 3 I would prefer for the slider value to be always visible; for example:

This would also fix 2 :+1:  So I'll see if that's easy.

antw:
This was merged yesterday; thanks for the fixes! :dancers:

PR #720, opened 2016-01-04
Issues/#703 speed up of les loads

grdw:


app/services/testing_ground/tree_sampler.rb
```diff
@@ -0,0 +1,27 @@
...
+
+        [carrier, GraphToTree.convert(graph)]
+      end]
+    end
+
+    private
+
+    def downsample(node_load, resolution)
+      if resolution == :low
+        node_load.each_slice(96).map(&:max)
```
grdw: @antw is it basically this what you were doing in the original implementation of the `downsampleCurve` or whas there more to it?

antw: Almost. The maximum *deviation* from zero should be selected. If you have a slice of two values [10, -15], then **-15** would be the one chosen.

grdw:
Also I have a question about:

https://github.com/quintel/etmoses/pull/720/files#diff-f5e04cf29632132df07eba5e7c34183bR100

This sort of implies that there are only 35040 / 52 = 673 points visible for a high resolution graph, right?

antw:
> This sort of implies that there are only 35040 / 52 = 673 points visible for a high resolution graph, right?

I believe so. That line [was originally added here](https://github.com/quintel/etmoses/commit/4270ac8f4718595703959af02ccba5f4532df0b3#diff-bad7651d16448a0f3d54747a54948cb7R12) and my understanding is that it tells `downsampleCurve` to return an array of 673 points, which is how many points are in a week.

No down-sampling is required in a high-resolution graph; I guess `downsampleCurve` is called there just to transform the array of values into an array of `[{x1, y1}, {x2, y2}, ...]`.

grdw:
Todo:

- [ ] Caching is broken. It looks as though the normal cache is also written for the strategies cache which results in a graph with the same charts (no difference in strategies). 
- [ ] When you select no strategies / zoom in / select a strategy there's a resolution bug.

grdw:
This is ready to be merged.

antw:
I was under the impression that requesting a high-res load curve would send only the currently selected node? For example, if I view an LES and select "HV 1" and then choose a weekly view, it would go to the server for the high-resolution curve for the "HV 1" node.

Instead it looks like the response contains the entire network with the high-resolution curves. Maybe I'm misunderstanding something, or perhaps both were tried and this method turned out to be better? (it actually is a bit better this way when debugging networks [with caching disabled](https://github.com/quintel/etmoses/blob/e48b295b4f004bf3b17b8612f6803dffe8c3abf0/config/settings/development.yml#L5))

grdw:
With caching disabled this is actually slower :cry: I didn't really think that there would be a situation where this would be practical without caching actually.

> I was under the impression that requesting a high-res load curve would send only the currently selected node? 

Well not exactly. It loads all the 'previously' selected nodes.

antw:
> Well not exactly. It loads all the 'previously' selected nodes.

Smart idea, but the responses I'm seeing definitely contain loads for the entire network even when the request just asks for one node. I'm perfectly happy for us to send all the high-res loads in this first version, but I suspect it's not what you intended.

A request was sent for `{ calculation: {resolution: "high", nodes: ["Battery"]} }` but what came back was JSON containing...

```json
{
  "networks": {
    "electricity": {
      "name": "HV Network",
      "load": [0.0, 0.1, "..."],
      "children": [{
        "name": "MV Network",
        "load": [0.1, 0.3, "..."],
        "children": ["..."]
      }]
    }
  }
}
```

app/services/testing_ground/tree_sampler.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        [carrier, GraphToTree.convert(graph)]
+      end]
+    end
+
+    private
+
+    def downsample(node_load, resolution)
+      if resolution == :low
+        node_load.each_slice(96).map do |loads|
```
antw: Instead of hard-coding 96, it would be better to determine this number based on the length of the load curve.

```ruby
class TreeSampler
  RESOLUTION_LENGTH_LOW = 365

  def downsample(*)
    if resolution == :low
      node_load.each_slice((node_load.length / RESOLUTION_LENGTH_LOW).floor)
      # ...
    end
  end
end
```

The reason for this being that the load curve length may not always be 35040; I have an LES locally which uses 8760 (one time-step per hour, instead of one per 15 minutes). This results in the low-res curves containing fewer values than expected.

```ruby
([0.0] * 35040).each_slice(96).map(&:sum) # => 365 values
([0.0] * 8760).each_slice(96).map(&:sum)  # => 96 values
```

app/services/testing_ground/tree_sampler.rb
```diff
@@ -0,0 +1,31 @@
...
+          absolutes = loads.map(&:abs)
+
+          loads[absolutes.index(absolutes.max)]
```
antw: This could be simplified as `loads.max_by(&:abs)`. [Enumerable#max_by](http://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-max_by).

grdw:
> Smart idea, but the responses I'm seeing definitely contain loads for the entire network even when the request just asks for one node. I'm perfectly happy for us to send all the high-res loads in this first version, but I suspect it's not what you intended.

I see what I did wrong. Moment please ;) 

Thanks for the comments! :+1: 
Will fix them asap.

app/services/network_cache/fetcher.rb
```diff
@@ -8,10 +8,12 @@ def self.from(testing_ground, opts = {})
 
     #
     # Fetches all cache and sets is as the load attribute for a node
-    def fetch
+    def fetch(nodes = [])
       tree_scope.each do |network|
         network.nodes.each do |node|
```
antw: We could avoid iterating through all the nodes in the network like so:

```ruby
nodes.map { |key| network.node(key) }.compact.each do |node|
  network.node(key).set(:load, read(network.carrier, node.key) || [])
end
```

Not a big deal, but it's slightly more efficient in large networks. The `map.compact` step is required in case the request specifies a node which does not exist. *Ideally* we'd reject the request (with `400 Bad Request` or `422 Unprocessable Entity`) before getting to this point, in which case the code becomes simpler:

```ruby
nodes.each do |key|
  network.node(key).set(:load, read(network.carrier, node.key) || [])
end
```

antw: ... the `map.compact` step means this is probably not worth changing unless we do that validation of the request (to assert that the nodes named actually exist). Your call, but I'm not too bothered if this doesn't get changed.

grdw:
> I see what I did wrong.

In other news, I can't.

With the following parameters:

`{"calculation"=>{"resolution"=>"high", "nodes"=>["Households 2", "Households 1"] } }`

I get the following result in the Javascript (and in the development.log):

thingy

It might be that you need to do a git pull since I put the node-specific things in at a later time than creating this pull request.

antw:
I bet it's because I have caching turned off. The Fetcher class sets the load on the nodes requested, but that will only kick in if caching is turned on. If it is turned off, the whole network will be returned.

I actually like that behaviour, since it avoids extra network calculations when caching is disabled. :+1:

app/services/network_cache/fetcher.rb
grdw: This is something which I have put in an [issue](https://github.com/quintel/etmoses/issues/722) for later. The thing is that the `TreeSampler` also contains a loop over all nodes which might than be simplified. And what to do when the nodes are empty? I'll put it in an issue and I'll figure this out later on. Thanks for this comment.

Issue #707, opened 2015-12-18
When saving 'as', the name should not be the same as the current LES


ChaelKruip:
Now it is possible to 'save as' with the same name which will lead to confusion.

grdw:
What would be prefered? The same as the topologies: (calling it `<name> - Clone #<x>`)?

ChaelKruip:
@JeroenvdLogt can you explain what you find confusing about the current situation?

![etmoses](https://cloud.githubusercontent.com/assets/1303760/11952020/bc31408e-a895-11e5-9483-b6584daeb57c.png)

ChaelKruip:
> What would be prefered?

I think it would be better to warn the user with a pop-up saying "Please choose a new name for your LES"

grdw:
That's also possible. We could also choose for something a little more subtle. We could track on a LES which LES is the original source. That way we can sort of say in a tiny sub-header "This LES is copied from <original LES name>"

JeroenvdLogt:
What is most important in opinion is that the user can keep track of the versions. My problem was (few weeks or more ago) that after changes in the LES the LES was saved as clone instead of under the same name. I cannot remember if this was my own fault or a temperory  bug in ET-Moses. 

To conclude, I think the most convenient solution woudl be:
Save the LES under the same name unless the user choose to save as. During save as user should be obliged by ET-Moses to choose another name (no cloning in the background without notice of the user).

I hope this will help. Otherwise you can call me.

ChaelKruip:
> I hope this will help. 

Very clear! Thanks!

grdw:
The easiest way to achieve is this is with a simple prompt pop-up that asks for a new name. 

![screen shot 2016-01-14 at 12 02 47 pm](https://cloud.githubusercontent.com/assets/2676542/12322802/be8176c8-bab6-11e5-8d5a-8201ebe59372.png)

'Annuleren'  and blanking the title will discontinue the saving process.

grdw:
![screen shot 2016-01-14 at 1 32 30 pm](https://cloud.githubusercontent.com/assets/2676542/12324505/49887cb0-bac3-11e5-8fbd-01deec558784.png)

Turned in it to someting a little more stylish.

JeroenvdLogt:
Great! I will test it..

PR #728, opened 2016-01-14
Adding a tiny prompt inbetween the saving process

grdw:
Fixes: #707

### DIFF
--- a/app/assets/javascripts/save_as.js
+++ b/app/assets/javascripts/save_as.js
@@ -1,17 +1,22 @@
-/*globals document*/
+/*globals document,customPrompt*/
 $(document).on('page:change', function () {
     'use strict';
 
     function submitSaveAsForm() {
         var form = $(this).parents("form");
         form.attr("action", $(this).data('url'));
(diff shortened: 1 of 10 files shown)


grdw:
Moment, spotted a problem.

app/assets/javascripts/save_as.js
```diff
@@ -1,17 +1,22 @@
...
         form.submit();
     }
 
     $(".save_as").on("click", function (e) {
         e.preventDefault();
 
-        window.saveAll.submitForms(submitSaveAsForm.bind(this));
+        var nameInput = $("input#testing_ground_name");
+
+        customPrompt("Please change the name of the LES", nameInput.val(), function (newName) {
```
antw: Might I suggest "Please set a name for your new LES" – that would make it clear you're setting the name of the *new* one. The current wording might imply that the user is changing the name of the *old* LES.

grdw: FIxed it :+1:

Issue #752, opened 2016-02-11
EV volume is off by a factor of 10000


ChaelKruip:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/12973466/d2e009ca-d0ac-11e5-8c13-2379a67b3533.png)

grdw:
Hmm, saw this happening yesterday aswell. The value we get from ETEngine is `0.0000096`

This is the literal value from et-engine:

````ruby
"storage"=>{"present"=>{"volume"=>0.02, "decay"=>4.4e-05}, "future"=>{"volume"=>0.02, "decay"=>4.4e-05}}
````

And in our builder we divide that number by 1000 to switch from Wh to kWh according to a comment.

antw:
The ETEngine value was changed from Wh to MWh (for consistency with the rest of the dataset), so to get kWh we now need to multiply by 1000.

grdw:
>  to get kWh we now need to multiply by 1000.

:+1:

PR #774, opened 2016-02-11
Fixes: #752

grdw:
See issues #752 

Nothing to fancy.


app/models/import/storage_volume_attribute.rb
```diff
@@ -1,6 +1,6 @@
 class Import
   StorageVolumeAttribute =
     Attribute.new('volume', 'storage.volume') do |value, *|
-      value / 1000 # Wh to kWh
+      value * 1000 # Wh to kWh
```
ChaelKruip: Shouldn't the comment be "# MWh to KWh"?

Issue #773, opened 2016-02-11
Wrong profile is shown in node overview


ChaelKruip:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/12978833/c3c86f78-d0d3-11e5-9380-1a2b708f3dd4.png)

but this is a group of 100 houses:

![etmoses](https://cloud.githubusercontent.com/assets/1303760/12978852/d46bd96e-d0d3-11e5-92dd-77540bca7bda.png)

ChaelKruip:
Actually, this the profile which is actually used! The EDSN profile shown in the technology table is never used!

ChaelKruip:
Including @antw

antw:
I can reproduce this by creating a scenario with 300 households, importing into Moses and deleting all technologies except the base load. The drop-down says "edsn_1a", but instead "anonimous_base_load_1" is saved.

If instead, when importing, I change the profile to "edsn_e1b" before saving, that profile *is* saved correctly. When I then try to edit the LES, "edsn_e1**a**" (the default) is selected in the profile `<select>`.

I suspect a bug in the technology editor table.

PR #777, opened 2016-02-11
Set's correct profile id + type for edsn base load

grdw:
Turns out that the old code responsible for updating the table was still present for EDSN. This is now updated to the correct code.

Fixes: #773

### DIFF
--- a/app/assets/javascripts/les/profile_matrix_editor/edsn_switch.js
+++ b/app/assets/javascripts/les/profile_matrix_editor/edsn_switch.js
@@ -5,19 +5,26 @@ var EdsnSwitch = (function () {
         validBaseLoads = /^(base_load|base_load_edsn)$/;
 
     function swapSelectBox() {
-        var technology   = $(this.target).data('type'),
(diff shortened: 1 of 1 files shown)


antw:
This solves the "anonimous" profile being selected, thanks!

I find that the Edit LES page is still defaulting to showing the first EDSN profile (e1a) in the `<select>` box, even though e1b. Do you know if this is a related problem?

screen shot 2016-02-11 at 16 00 54
screen shot 2016-02-11 at 16 00 59

grdw:
> Do you know if this is a related problem?

Yes it is. I'll look into it

grdw:
It's a little more complicated than it sounds. I also remember this is not the first time this problem bites us in the butt. So I think there needs to be a more 'simple' fix for this problem. I'll make an issue of it. The current fix is working - I checked a lot of things (switching, change units, change units and switching, saving, switching ,changing units in all possible orders you can imagine but feel free to check). :sweat_smile:

PR #779, opened 2016-02-11
Set profile once <select> has been initialized

antw:
I hate to say it, but I tried importing the same 300 household scenario again and, despite the import form showing "edsn_e1a", the base load got saved with "anonimous_base_load_1".

A previous version of the pull request #773 (2ba2ed5) did work, so I resurrected one of the lines which was removed from the merged version.

@grdw Does this look right to you?


grdw:
Yes, I'm still convinced that this needs to change from the root down but for now this will do.

PR #782, opened 2016-02-12
Issues/#759 maximising concurrency should not redistribute technologies over end points

grdw:


antw:
I might be misunderstanding, but the concurrency settings don't seem to work correctly with this PR.

1. I added 3 units of "Electric Vehicle" to an endpoint.
2. Changed EV concurrency to "Min".
3. Clicked "Apply".
4. All three units of EVs are still grouped together.

I thought this should split the EVs into three separate technologies (attached to the same endpoint as before) with randomly-assigned profiles? On `master` this *does* split the EV into three separate parts, but distributes them across all the endpoints.

Correct me if I'm wrong please!

grdw:
> Correct me if I'm wrong please!

Hmm.. good question. I'll write an extra spec for this to validate the outcome. I'm still wondering what the intention should be when you do that to be honoust.

grdw:
Was a little tiny bug in the Javascript though. And you're right. I think it should be workig now.

ChaelKruip:
> I'm still wondering what the intention should be when you do that to be honest.

Can you please explain what is unclear to you @grdw? Perhaps we need to write up our expectations somewhere very clearly (also as part of the user documentation) and make all appropriate specs to settle this (recurring) issue once and for all.

grdw:
> Perhaps we need to write up our expectations somewhere very clearly

Yes, we should. However, the problem described was already clear but I had to rethink it again. Maybe we should still write it in the user documentation though.

antw:
> Perhaps we need to write up our expectations somewhere very clearly (also as part of the user documentation) and make all appropriate specs to settle this (recurring) issue once and for all.

I think this would be good, because it's not clear to me what the *correct* behaviour is. Here's what I just tried locally:

1. Topology with three endpoints (Households 1, Households 2, and Households 3).
2. Households 1 has a base load technology with 1 unit.
3. Households 2 has a single EV technology with 3 units.
4. Households 3 has no technologies attached.

If I choose to **minimise EV concurrency**, I find that:

* Households 1 now has a base load and two separate EV technologies with 1 unit each.
* Households 2 has a single EV technology with 1 unit.
* Households 3 has nothing *(correct)*

Two units of EV have moved from Households 2 to Households 1. This *may* be intentional; it doesn't feel right to me, but if it's intentional then so be it.

If I now choose to **maximise EV concurrency** I find:

* Households 1 has a base load and a single EV with 3 units
* Households 2 has nothing
* Households 3 has nothing *(correct)*

I think some documentation would be good, so that we can describe exactly how technologies should be distributed when changing the concurrency settings.

ChaelKruip:
Following @antw example I would expect the following behaviour:

If I choose to **minimise EV concurrency**, I find that:

* Households 1 still only has a base load
* Households 2 has three individual EVs with 1 unit each
* Households 3 has nothing *(correct)*

If I now choose to **maximise EV concurrency** I find:

* Households 1 has a base load (still)
* Households 2 a single EV technology with 3 units
* Households 3 has nothing *(correct)*

grdw:
I wrote a spec for this situation and I can confirm this behavior. :+1: Also fixed it though.. because it was only adding two characters in the existing code to actually fix it.

grdw:
Maybe abuse the Wiki for a good tough article about concurrency: https://github.com/quintel/etmoses/wiki/Concurrency ?

ChaelKruip:
> Maybe abuse the Wiki for a good tough article about concurrency:

I'll write up something!

antw:
Works nicely, thanks again @grdw! :cake:

PR #792, opened 2016-02-16
Minor bug fixes to profile matrix editor

grdw:
- Show input capacity in front end as a disabled box (show the output of the calculation)
- Spotted a minor bug with the position relative to buffer radio boxes
- Cleaning up attributes of `InstalledTechnology` class.

### DIFF
--- a/app/assets/javascripts/extensions.js
+++ b/app/assets/javascripts/extensions.js
@@ -54,7 +54,7 @@ $.fn.extend({
     set: function (option, value) {
         'use strict';
 
-        $(this).data(option, value)
+        $(this).data(option.camelize(), value)
                .attr("data-" + option.underscorize().replace(/\_/g, '-'), value);
     }
 });
@@ -66,3 +66,11 @@ String.prototype.underscorize = function () {
         return "_" + a.toLowerCase();
     });
 };
+
+String.prototype.camelize = function () {
+    'use strict';
+
+    return this.replace(/[-_\s]+(.)?/g, function (_, c) {
+      return c ? c.toUpperCase() : '';
+    });
+}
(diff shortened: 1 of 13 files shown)


antw:
* A couple of technologies (power-to-gas, neighbourhood battery) give this when I add them to an (existing) LES:

  screen shot 2016-02-17 at 14 18 32

* I think it would be nice to put the input and output capacity fields next to each other; either on the same line, or one on top of the other. At the moment they don't appear to be related, and you have to scan diagonally across the page if you want to compare the two:

  screen shot 2016-02-17 at 14 19 56

  Perhaps we could move "output capacity" to be where "demand" is currently located, move "input capacity" to where "output capacity" is, and move "demand" to the second line?

  screen shot 2016-02-17 at 14 22 38

  ... or:

  screen shot 2016-02-17 at 14 28 48

* The field name "Input capacity (electricity) [kW]" is not always true: the gas component of the hybrid heat-pump takes *gas* as the input. If the technology editor has no way of knowing the input carrier, I suggest we just remove "(electricity)" from the title.

grdw:
> ... or:

I like the second one better though :+1: It's easier to see for the eye.

> the gas component of the hybrid heat-pump takes gas as the input.

We're adding the hybrid heat pump as two separate technologies now, so we do know the carrier if I'm not misaken?

antw:
> We're adding the hybrid heat pump as two separate technologies now, so we do know the carrier if I'm not misaken?

We do know, and the database/Ruby knows, but I'm not sure if the technology editor has this information? If so, great: we can swap out "electricity" for "gas". If it doesn't, it's not a disaster if we have to remove the carrier name.

Issue #794, opened 2016-02-17
Distinguish between automatically selectable profiles


ChaelKruip:
As mentioned in https://github.com/quintel/etmoses/issues/784, only profiles that are 'approved' should be eligible for selection by ETMoses.

grdw:
I disagree with the term 'approved' here. It almost sounds if every load profile that's been uploaded is wrong by default. Also who can set the 'approved' checkbox? Only admins I would guess. 

You don't want the profile to take part in the concurrency process. So I would guess it's something like `can_concur` or `included_in_concurrency`. Or will we use the `approved` boolean in the future for something else besides the fact that it can or cannot be selected while minimizing the concurrency?

ChaelKruip:
> included_in_concurrency

Fine by me!

PR #797, opened 2016-02-17
Issues/#794 distinguish between selectable profiles

grdw:
Please merge https://github.com/quintel/etmoses/pull/782 first - before merging this.

---

This PR closes: #794 

Adds a boolean to `load_profiles` that includes/excluded load profiles from the concurrency method.

### DIFF
--- a/app/controllers/load_profiles_controller.rb
+++ b/app/controllers/load_profiles_controller.rb
@@ -63,7 +63,7 @@ class LoadProfilesController < ResourceController
   def profile_params
     params.require(:load_profile).permit(
       :key, :name, :public, :load_profile_category_id, :default_capacity,
-      :default_demand, :default_volume,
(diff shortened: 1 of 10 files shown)


antw:
I made two small changes: [use the LoadProfilePolicy instead of checking `user.admin?` directly](https://github.com/quintel/etmoses/commit/5eb6988ae8098032d96ea15e67b6a15272bde43b) and [ignore `:included_in_concurrency` from unauthorised users](https://github.com/quintel/etmoses/commit/4e7dc9da557f30e43057ed464a65a464e6546c19).

The second change is important, since hiding a form field isn't enough to stop a malicious user from sending a raw HTTP request and changing the value. You have to actually fail the request, or discard the value which the user is not authorised to change.

Thank you though, this does work perfectly. :+1:

PR #815, opened 2016-02-23
Replace the NVD3 chart with a regular D3 chart

grdw:
- Also includes a line with a hover


antw:
@grdw Is this ready or a work-in-progress? I found a couple of issues, but I'll hold off on posting them if you're still working on it.

grdw:
> @grdw Is this ready or a work-in-progress?

Well you could post them because I probably haven't found them yet :angel:

antw:
> Well you could post them because I probably haven't found them yet

I wonder if I'm doing something wrong; I'm running c152595 with caching disabled (I don't think that should matter) and have quite a few problems:

--

While waiting for the calculation, the week-selector is visible:

screen shot 2016-02-25 at 12 59 40

--

The "brush" is just a grey bar and doesn't include the data (maybe this is intentional, but it feels like a step backwards to me):

screen shot 2016-02-25 at 13 03 14

--

The annual view says that the first date is 1970:

screen shot 2016-02-25 at 13 02 16

--

Choosing to view data for one week appears not to load the high res data (no HTTP request is sent), and the dates/times are wrong:

screen shot 2016-02-25 at 13 08 25

--

The value "0" should always be visible on the vertical axis, but often isn't.

01 wrong-dates no-high-res

--

The hover line (which I really like) shows in in the weekly view, but the value doesn't:

| Annual view | Weekly view |
|---|---|
| screen shot 2016-02-25 at 13 10 48 | screen shot 2016-02-25 at 13 10 59 |

--

Selecting a very small range in the "brush" results in no data being shown (this was also a problem with NVD3):

screen shot 2016-02-25 at 13 15 21

--

I'd quite like it if we could take the font-size on the down to 13px or 12px. The default 14px feels a bit big.

--

Things I like:

* The chart feels more responsive.
* I love that it can display the *time* of the data and not just the day; that's really nice and will help when debugging the calculation.
* The "hover" line is much better than what NVD3 did, where you had to hover over the data.
* The chart no longer displays Jan 01 on the right-hand side of the horizontal axis. That was always confusing, and this is much better.

grdw:
@antw thanks for the points! I'll look into them. I must say that I've tested with caching enabled ... which is probably a reason for different results. But I'll turn it off to see what happens. :+1:

grdw:
### Fixed:

> While waiting for the calculation, the week-selector is visible:

Done! :smile:

---

### Need to checkout:

> The "brush" is just a grey bar and doesn't include the data (maybe this is intentional, but it feels like a step backwards to me):

Yes... I don't really feel strong about it but I can see if I can make it return.

> The value "0" should always be visible on the vertical axis, but often isn't.

There is still a thing with the max and min values of the chart. Those are not present in the current chart.

--- 

### Can't reproduce:

> The annual view says that the first date is 1970:

I can't reproduce this one to be honest? Do you have to select a specific week? I can remove the formatting for a 'year' since it doesn't make sense anyway to include it.

> The hover line (which I really like) shows in in the weekly view, but the value doesn't:

This probably has to do with the 'wrong' resolution but I can't really reproduce this locally:

screen shot 2016-02-25 at 2 39 28 pm

---

### Minor problem:

> Selecting a very small range in the "brush" results in no data being shown (this was also a problem with NVD3):

Hmmm yes. We could maybe clear the brush once it's too small? However what is 'too small'?

antw:
FYI The latest commit 53d8fee doesn't work in Chrome. The `tickFormat` on `xAxis2` seems to prevent the chart from rendering:

"Uncaught TypeError: Cannot read property '1' of undefined"

screen shot 2016-02-25 at 15 24 39

It does render if I remove the tickFormat ([d3_load_chart.js:454](https://github.com/quintel/etmoses/pull/815/files#diff-13a62fd4affca3b7a9a61a5897723853R454)) but of course that messes up the axis formatting.

It does render in Firefox, but the values (and brush) don't look right:

screen shot 2016-02-25 at 15 27 21

I cleared my localSettings just in case something from the old chart was interfering, but that has no effect.

grdw:
> "Uncaught TypeError: Cannot read property '1' of undefined"

I had a weird bug with this line just yet, It had to do with the year formatting line I removed earlier... the 1970 you showed me. The plot thickens ...

grdw:
Chrome:
screen shot 2016-02-25 at 5 12 44 pm

Firefox:
screen shot 2016-02-25 at 5 12 59 pm

I can't get the brush that 'ugly'. I do know that there's some padding problem due to the fact that it doesn't reduce the height by the tickHeight.

grdw:
I fixed the high resolution problem. It was appearantly only the case for LES's without strategies. I fixed that problem in the current commit. Short question though? Should I always rebase and force push in these cases or create loose commits and when you say "ok" rebase and force push? :sweat_smile:

antw:
> Short question though: Should I always rebase and force push in these cases or create loose commits and when you say "ok" rebase and force push?

Which would work better for you? Separate commits would be useful for seeing when changes have happened, but having to wait for an "okay" before rebasing might get annoying for you.

grdw:
> Which would work better for you? 

I'm okay with it the way it is right now, but I thought it might be annoying for you.

> Separate commits would be useful for seeing when changes have happened,

Yes, you can also see it when the 'one' squashed commit is at the bottom of the conversation of a PR.

antw:
> > The annual view says that the first date is 1970:
>
> I can't reproduce this one to be honest? Do you have to select a specific week? I can remove the formatting for a 'year' since it doesn't make sense anyway to include it.

I was able to fix this by removing the "year" formatting from the chart (and pushed another commit to make all the chart lines pixel-perfect).

There are some weird things going on with the x-axis formatting in Firefox though:

| Browser | Image |
|---|---|
| Firefox | screen shot 2016-02-29 at 13 07 46 |
| Chrome | screen shot 2016-02-29 at 13 09 34 |

I'm not sure what's going on here. :confused:

I think I would be inclined to rename the various `*-load-graph` CSS classes to `*-load-chart` so as not to be confused with the graph (network) data structure.

Firefox formatting aside, I think this is otherwise perfect, and a big improvement over the old chart. :+1:

grdw:
> I was able to fix this by removing the "year" formatting from the chart (and pushed another commit to make all the chart lines pixel-perfect).

I also toggled between the two but sometimes it also produced errors in the console ... It has to do with the multi-format of d3, that's certain. I'll investigate a little to see what might cause this weird behavior.

grdw:
> I'm not sure what's going on here. :confused:

Isn't this a Firefox version thingy? 

Chrome (48.0.2564):
screen shot 2016-02-29 at 2 23 57 pm

Firefox (44.0.2):
screen shot 2016-02-29 at 2 24 02 pm

They're almost identical, except the brush is a little bit smaller/bigger due to my not so steady hand.

antw:
> Isn't this a Firefox version thingy?

Sadly no, I also have 44.0.2. I'll try to figure it out since you can't reproduce it.

antw:
> I'll try to figure it out since you can't reproduce it.

This is thoroughly confusing. :confounded: 

I do think I've managed to figure out what triggers it though: my timezone is GMT (UTC+0). If I set it to CET (UTC+1) the problem goes away.

All the dates created by the D3LoadChart are consistent (and correct) in Chrome and Firefox, but by the time it gets to the d3.time.format.multi ([here](https://github.com/quintel/etmoses/pull/815/files#diff-13a62fd4affca3b7a9a61a5897723853R43)) the Firefox dates are all off by an hour.

grdw:
> I do think I've managed to figure out what triggers it though: my timezone is GMT (UTC+0)

Cool! I see it to. I'll see what I can manage though :+1:

grdw:
I can also only reproduce it for BST (British summer time). I now also see the other date problems.. damn. Just googling 'new Date firefox chrome' returns more horror than needed.

grdw:
I fixed the time zone problem! :dancers: 

See: https://github.com/quintel/etmoses/commit/63834314e580099cd6b8a5e198a6fa7447e760e0

grdw:
There's still one minor difference between Chrome & Firefox. If you are in between two days in the year view. Firefox formats it to "13:00"

screen shot 2016-03-02 at 09 58 41

And Chrome format's this at "12:00". 

screen shot 2016-03-02 at 09 58 48

antw:
Nice! Load charts look perfect! :cake: 

> There's still one minor difference between Chrome & Firefox. If you are in between two days in the year view. Firefox formats it to "13:00"

I can fix that one by changing `d3.time.format.multi` to `d3.time.format.utc.multi`

--

I did discover a couple of *minor* issues when viewing the load profiles/price curves.

* Negative values in the EV load profiles are not shown in the main chart (but are in the smaller "brush" chart). See /load_profiles/287.

* Unit formatting on the charts has been lost, and all show "kW". For example in the financial curves (/profiles/1 which should show `€`) or the behaviour curve (/profiles/11 which should be `On/Off`).

* The hover line isn't available on the load profiles/price curves. Not a deal-breaker, but it's a pity since it's such a nice feature.

I don't think any of these are a huge deal. @grdw if you don't have time to investigate, we can merge this and come back to these issues later?

grdw:
I have some time left before lunchbreak so I think I'll just pick these one's up than continue with the calculation per week. :+1:  It's not to much work.

grdw:
FYI:

> Negative values in the EV load profiles are not shown in the main chart (but are in the smaller "brush" chart). See /load_profiles/287.
> Unit formatting on the charts has been lost, and all show "kW". For example in the financial curves (/profiles/1 which should show €) or the behaviour curve (/profiles/11 which should be On/Off).
> The hover line isn't available on the load profiles/price curves. Not a deal-breaker, but it's a pity since it's such a nice feature.

Are done. There's only one thing, the price curves and behavior profiles are loaded in high resolution which is not a major problem but it makes brushing a little slow.

Issue #823, opened 2016-02-29
Yearly investment of a business case returns NaN because full load hours is 0.


grdw:
See title.

This causes the business case to calculate incorrectly for it will ignore all other investment attributes and returns NaN.

PR #824, opened 2016-02-29
Fix NaN bug for initial costs of business cases

grdw:
Fixes: #823

### DIFF
--- a/app/models/installed_technology.rb
+++ b/app/models/installed_technology.rb
@@ -213,11 +213,19 @@ class InstalledTechnology
     super.merge('carrier_capacity' => carrier_capacity)
   end
 
-  def yearly_investment
-    om_costs_per_year.to_f                   +
-    initial_investment.to_f                  / (technical_lifetime || 1) +
-    om_costs_per_full_load_hour.to_f         / (full_load_hours || 1) +
-    om_costs_for_ccs_per_full_load_hour.to_f / (full_load_hours || 1)
+  def total_yearly_costs
+    (depreciation_costs     +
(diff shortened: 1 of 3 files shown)


app/models/installed_technology.rb
```diff
@@ -214,10 +214,18 @@ def as_json(*)
...
-    om_costs_for_ccs_per_full_load_hour.to_f / (full_load_hours || 1)
+    (initial_investment.to_f / (technical_lifetime || 1)) + yearly_om_cost
+  end
+
+  def yearly_om_cost
+    om_costs_per_year.to_f + (
+      om_costs_per_full_load_hour.to_f +
+      om_costs_for_ccs_per_full_load_hour.to_f
+    ) / yearly_full_load_hours
+  end
```
grdw: @antw not really sure about this code formatting. What is prefered in these situations?

antw: I think the solution you ended up with – moving variable costs to a separate method – was the nicest; it avoids having to think about which value is being divided by `yearly_full_load_hours`.

Maybe something like this would also have been okay (all it does is move the opening bracket down a line to help emphasise that the fixed cost is not divided by the full load hours):

```ruby
def yearly_om_cost
  om_costs_per_year.to_f +
    ( om_costs_per_full_load_hour.to_f +
      om_costs_for_ccs_per_full_load_hour.to_f
    ) / yearly_full_load_hours
end
```

The original version passes [Rubocop](https://github.com/bbatsov/rubocop), so I think it's just personal preference. :smiley:

PR #831, opened 2016-03-03
Check if the actual accompanied delayed job is present when recalculating a LES

grdw:
Thing is now that when somebody calculates a LES the first time, the cache is properly written but somebody can still cancel the request (for instance out of a rush, before a demo, moving away from the page) (which results in the `testing_ground_delayed_job` not being destroyed).

The next time somebody makes a change the calculator sees that there already was a previous calculation (`testing_ground_delayed_job`) present and finished, which will cause the server to think that it's done while there's nothing valid in the cache is to show to the front-end (which is the error Alex was seeing).

So my fix would be when recalculating to check if the accompanied `delayed_job` of the `testing_ground_delayed_job` didn't already finish (i.e. if it's still present or not).

### DIFF
--- a/app/assets/javascripts/les/topology_load_charts/d3_load_chart.js
+++ b/app/assets/javascripts/les/topology_load_charts/d3_load_chart.js
@@ -520,7 +520,6 @@ var D3LoadChart = (function () {
                 .attr("transform", "translate(0," + height + ")")
                 .call(xAxis);
 
-            console.log(data.type);
             svg.append("g")
                 .attr("class", "y axis")
                 .call(yAxis)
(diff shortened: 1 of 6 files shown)


antw:
Out of interest, what's the purpose of the `finished_at` column? Is there a valid case where a calculation has finished, but we want to keep the `TestingGroundDelayedJob` record?

If not, what are your thoughts on getting rid of the `finished_at` column and instead deleting the record [when the calculation has finished](https://github.com/quintel/etmoses/blob/7fef9f2a408a510b8687d39d6aff0f515c7a8cf8/app/jobs/testing_ground_calculator_job.rb#L15)?

grdw:
> Out of interest, what's the purpose of the finished_at column?

It could just aswell be a boolean called finished. But because of a tutorial it sort of pointed me towards this solution.

> Is there a valid case where a calculation has finished, but we want to keep the TestingGroundDelayedJob record?

When the calculation is pending, otherwise no.

> If not, what are your thoughts on getting rid of the finished_at column and instead deleting the record when the calculation has finished?

Sounds good :+1: I'll fiddle a little.

grdw:
> If not, what are your thoughts on getting rid of the finished_at column and instead deleting the record when the calculation has finished?

This actually works pretty well :sweat_smile:

antw:
:+1:

Issue #695, opened 2015-12-15
Changing the name of a LES should not recalculate your business case


grdw:
*It's all in the title.*

It's a bit unnecessary to fully recalculate the business case when changing the name of the LES.

ChaelKruip:
Putting this on-hold as it is not urgent

Issue #828, opened 2016-03-01
[ Technical ] It should be possible to run the load calculations for time-intervals


grdw:
Part of #802 

----

**Rough draft**

*Current situation:*
Currently the load calculation is done for the full year and all points are stored in cache. Afterwards the content of the nodes are fetched from the cache on a resolution base.

I would suggest to split the cache up into separate parts.

Right now the following folder structure is present:
````
<environment>
│   ├── <les id>
│   │   └── basic
│   │       ├── electricity
│   │       └── gas
````

I'd suggest to change it to:
````
<environment>
│   └── 442
│       └── basic
│           ├── electricity
│           │   ├── 0
│           │   └── 1
│           └── gas
│               ├── 0
│               └── 1
````
Where the numbers are the week numbers of the selection. The week 0 will always contain a low resolution calculation. 

---

**Idea 1: cut the load profiles when needed**

Technically I would think - and this is purely speculative because I'm sort of lost in the woods when looking at the calculation code - that it's giving the [calculation context](https://github.com/quintel/etmoses/blob/master/app/models/calculation/context.rb) an extra option called week number.

Based on the week number you can determine the resolution (week 0 = low resolution, week 1 till 52 is high resolution). Based on the week number you can also determine which set of frames to calculate (in essence `max = (length / 52) * week_number, min = (length / 52) * (week_number - 1)`. 

*What if week 0?*
Week 0 means the resolution is low. This will mean that we can group the load profile values by the max load of a certain time-step (the max of every 96th point if the length of the load profile is 35040) Like we are doing now but after the calculation is finished.

*What if week 1-52?*
Week 1-52 means the resolution is high. This will mean that we simply take the load profile values but cut a range out.  Like we are doing now but after the calculation is finished.

The only thing I'm worrying about this simple setup I have here is - what will happen at the last frame of a week when a strategy is applied? Does that need to be calculated for the whole year like it is now and than cut out in order to be effective? Or some check to see if the week number is not 52 or 0? 

*In short:*
Where I'm leaning towards is an option where the load profiles are pre-cut based on a resolution and a range of frames. However I have no idea if that has bigger implications considering strategies.

---

**Idea 2: Pre-cut the load profiles when uploading them**
[...]

Issue #802, opened 2016-02-18
It should be possible to run the load calculations for time-intervals


ChaelKruip:
Performance is the main reasons why this is a much wanted feature:

Most of the time, people will zoom in on a particular time-interval (say a week) and play with the various strategies and technologies. Calculating the whole year makes this process frustratingly slow. This will become more so if we add load-per-technology and a smarter load calculation (subnet awareness).

## Short-term requirements

For the current sprint, it suffices if the user can use the current year/week selector to choose the period of calculation. 

When you first open the overview page something like this would show:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/13143615/662d80b6-d645-11e5-8c1c-fe3189c8b34c.png)

### Calculating the whole year

If you select the 'whole year' option, the whole year is calculated and the app works as it currently does (zooming in on a week works the same).

### Calculating a week

If you select a specific week, the load is only calculated for that week. All other weeks are not available (yet) until you select them to be the time-interval for calculation.

Business cases are not available until you calculate the whole year.

### Behaviour upon editing

If you edit an aspect of the LES, the load is recalculated for the currently selected time-interval (even if you previously calculated the whole year, the current week is re-calculated if you are in week-mode).

### Initial conditions

Calculating for shorter time-periods raises the question: what is the state of buffers and storage devices? They typically depend on the past so what happened before?

I propose to make the following assumption: EV, P2P and buffers are half-full to begin with. This is assuming the least about the past and leaves flexibility open in two directions.

@jorisberkhout @dennisschoenmakers @AlexanderWirtz do you think this is a good starting point as well?

## Long-term requirements

After discussion with @dennisschoenmakers I've realised that in the future time-intervals of different duration and (a single point, a day, a month) or multiple intervals (a series of days spread over the year etc.) might become relevant to show as well. It would be good if this is kept in mind when re-designing the load calculation. Perhaps the load function could be given an array of start/stop times as an argument?

AlexanderWirtz:
I like it

grdw:
> If you select the 'whole year' option, the whole year is calculated and the app works as it currently does (zooming in on a week works the same).

It shouldn't in my opinion. I think there should be something shown to the user. If we could calculate the *whole topology* in a certain resolution that would save time I would guess (if it's possilble). I think that calculating something on a per node basis might also be a possibility but I can't say for sure due to the questions that are also still present at #828. 

If I get it correctly this is what happens (@antw please hit me if I explain it badly):

A topology:

[...]

Issue #768, opened 2016-02-11
Editing the market model or business case should not trigger a new load calculation


ChaelKruip:
Perhaps the cache is invalidated too conservatively?

grdw:
Linked: https://github.com/quintel/etmoses/issues/695

ChaelKruip:
Putting this on-hold as it is not urgent

ChaelKruip:
Putting this on-hold as it is not urgent

grdw:
I just tested this but it doesn't really work that well.. whenever the `markt_model_id` changes (for instance, you make a change to a market model that isn't yours and it starts cloning) than it still triggers a recalculation.

Also the `save all and view Les` button triggers a save action on the LES itself which causes the cache to flush.

grdw:
Closing, just tested on beta and this is fixed.

PR #837, opened 2016-03-08
Issues/#828 calculate loads per week

grdw:
@antw could you test this out? 

---

Fixes: #695, #768, #802, #828


antw:
I'm not having much joy testing this locally. Apparently [time_frame](https://github.com/quintel/etmoses/blob/6593659195e6390ef3e57ba9631e3a60b2877b45/app/services/network_cache/cache_helper.rb#L28) is nil, which causes:

```
no implicit conversion of nil into String
.rbenv/versions/2.3.0/lib/ruby/2.3.0/pathname.rb:409:in `initialize'
.rbenv/versions/2.3.0/lib/ruby/2.3.0/pathname.rb:409:in `new'
.rbenv/versions/2.3.0/lib/ruby/2.3.0/pathname.rb:409:in `join'
moses/app/services/network_cache/cache_helper.rb:29:in `file_name'
moses/app/services/network_cache/validator.rb:20:in `block (2 levels) in cache_intact?'
moses/app/services/network_cache/validator.rb:19:in `each'
moses/app/services/network_cache/validator.rb:19:in `all?'
moses/app/services/network_cache/validator.rb:19:in `block in cache_intact?'
...
```

This happens when testing an existing scenario, with caching on and off.

grdw:
Do you open an old LES or create a new one? Also is the `load` statically set in that LES (because I know some LES's you have created sometimes have that feature). It's hard to imagine that `time_frame` is nil. The only thing that can cause that is if your `resolution` is neither 'high' or 'low'. Maybe it's blank (`''`)? 

````ruby
[11] pry(main)> NetworkCache::Validator.new(TestingGround.last, resolution: '').time_frame
  TestingGround Load (0.4ms)  SELECT  `testing_grounds`.* FROM `testing_grounds`   ORDER BY `testing_grounds`.`id` DESC LIMIT 1
=> nil
```
However that's also a little hard to get into since it defaults to 'high':

The resolution for the testing ground calculator is always set to ['high'](https://github.com/quintel/etmoses/blob/issues/%23828-calculate-loads-per-week/app/services/testing_ground/calculator.rb#L72
) or something the user specified. Also it defaults to 'high' in the [Javascript](https://github.com/quintel/etmoses/blob/issues/%23828-calculate-loads-per-week/app/assets/javascripts/les/topology_load_charts/d3_load_chart.js#L347).

So I'm actually kind of curious what's causing this behavior? :sweat_smile:

antw:
It was happening with every LES I tested. I checked out the branch again and now I can't reproduce it. Never mind. :confounded:

antw:
The basic functionality works well, and it's great to see how much faster the UI feels when we only have to calculate a week. I did notice one fairly big problem however: `Network::Curve` has two attributes – [`resolution` and `frames_per_hour`](https://github.com/quintel/etmoses/blob/3d4944359cddb9d66bd0995b6175a22515de08fb/app/models/network/curve.rb#L26) – which are based on the length of the curve. These attributes are used in a couple of places in the load calculation so that we can correctly convert between kW and kWh.

Because the weekly profiles contain only a 672 elements instead of 35040, the `resolution` is wrong; thus the calculation of any technology with a `volume` attribute is incorrect.

I did start to put together [a class to preserve the attributes of the original curve](https://gist.github.com/antw/6748168bf0303a64f196), but had to deal with other tasks before finishing. Sadly this class alone is not enough to solve the problem:

* There are a couple of places where we call `Network::Curve.new` which should be replaced with `Network::Curve.from`

* [Calling `map`, `reduce`, or any other `Enumerable` method](https://github.com/quintel/etmoses/blob/3d4944359cddb9d66bd0995b6175a22515de08fb/lib/paperclip/scaled_curve.rb#L23-L24) causes the `Curve` to become an `Array`, and we lose the `resolution` information.

An alternative is to have the `Network::Context` calculate (or be told) the resolution, and pass it as an option to each technology ([here](https://github.com/quintel/etmoses/blob/3d4944359cddb9d66bd0995b6175a22515de08fb/app/models/calculation/technology_load.rb#L28-L43)).

I'm also open to any other ideas.

--

Here's an example of how the incorrect `resolution` messes up loads. This is an endpoint with an ordinary P2P battery, and an "Other" technology with a capacity of -10 (it produces a constant 10kW).

**Before:**

screen shot 2016-03-21 at 09 31 30

**After:**

screen shot 2016-03-21 at 09 31 24

(ignore the different axis, that has been fixed on master already).

It would be nice to have before and after shots like this for other technologies to ensure nothing else has been broken; I'm especially curious if [the calculation of the ratio](https://github.com/quintel/etmoses/blob/3d4944359cddb9d66bd0995b6175a22515de08fb/app/models/curve_collection.rb#L38-L40) (for base loads) is still correct when only using a week of data. I suspect not, since profiles used for "Postponing of individual base load" [do not have a fixed ratio](https://beta-moses.energytransitionmodel.com/load_profiles/311) whereas [the EDSN profiles](https://beta-moses.energytransitionmodel.com/load_profiles/240) do..

--

Finally, [profile_length](https://github.com/quintel/etmoses/blob/issues/%23828-calculate-loads-per-week/app/models/installed_technology/profile_curve.rb#L35-L41) is not really an attribute of the technology. I think it would be nice to find another way to do this; I think a CurveSlice class could remember the length of the original profile, which might avoid the need to add this method to `InstalledTechnology`.

grdw:
It sounds a lot like there are loads of specs missing. :sweat: 

Thanks for your feedback.

grdw:
> Because the weekly profiles contain only a 672 elements instead of 35040, the resolution is wrong; thus the calculation of any technology with a volume attribute is incorrect.

I don't really get this. Thinking out loud: the resolution was either 1 or 0.25 (8760.0 / 35040.0) but now it becomes 13.0357 because the `length` attribute is 672 which causes a bad calculation. However, we expect it to be 0.25 since the *resolution of the chart* is high, am I right?

grdw:
screen shot 2016-03-21 at 12 30 29

Is it this detail that's missing in the other chart?

antw:
> Is it this detail that's missing in the other chart?

Yes.

> Thinking out loud: the resolution was either 1 or 0.25 (8760.0 / 35040.0) but now it becomes 13.0357 because the length attribute is 672 which causes a bad calculation. However, we expect it to be 0.25 since the resolution of the chart is high, am I right?

Yes, this is correct. However it's not the resolution of the *chart*, but the resolution of the *profile*. A profile with 8760 values has a resolution of 1.0 (and `frames_per_hour` = 1). A profile with 35040 values has a resolution of 0.25 (`frames_per_hour` = 4).

This is used in a couple of places in the calculation, but most notably [here when setting the volume of a technology](https://github.com/quintel/etmoses/blob/3d4944359cddb9d66bd0995b6175a22515de08fb/app/models/network/technologies/generic.rb#L108), which allows easy conversion between loads received by a technology in kW, and the energy which is stored, in kWh.

Basically, if the original full-length profile has a resolution of 0.25, the weekly profile – containing 672 values – should also report that its resolution is 0.25 (and `frames_per_hour` = 4).

grdw:
> Basically, if the original full-length profile has a resolution of 0.25, the weekly profile – containing 672 values – should also report that its resolution is 0.25

Stating that length can now by 672 or 35040 
Can I be so bold to just derrive the resolution from the length like:

````ruby
HIGH_RESOLUTION = 0.25
LOW_RESOLUTION = 1.0

def resolution
  case length
  when 673, 35040
    HIGH_RESOLUTION
  else
    LOW_RESOLUTION
  end
end
````

**Comment by antw** [comment]
> Can I be so bold to just derrive the resolution from the length like:

I suppose so, yes!

**Comment by grdw** [comment]
> It would be nice to have before and after shots like this for other technologies to ensure nothing else has been broken; I'm especially curious if the calculation of the ratio (for base loads) is still correct when only using a week of data. I suspect not, since profiles used for "Postponing of individual base load" do not have a fixed ratio whereas the EDSN profiles do..

- [ ] Test + fix

> Finally, profile_length is not really an attribute of the technology. I think it would be nice to find another way to do this; I think a CurveSlice class could remember the length of the original profile, which might avoid the need to add this method to InstalledTechnology.

- [ ] Test + fix

**Comment by grdw** [comment]
> Postponing of individual base load

In light of this problem I created a spec calculating the first week of a LES with only a single base load technology (which is congested). The issue that arrises here is that it correctly skips the 3 hours, but eventually at the last point of the week it will dump all that it skipped (if the rest of the trail is congested).

**Comment by grdw** [comment]
These are the specs I'm talking about currently: [here](https://github.com/quintel/etmoses/blob/issues/%23828-calculate-loads-per-week/spec/controllers/testing_grounds/calculation_spec.rb#L153-L164)

**Comment by antw** [comment]
> The issue that arrises here is that it correctly skips the 3 hours, but eventually at the last point of the week it will dump all that it skipped (if the rest of the trail is congested).

[That occurs here](https://github.com/quintel/etmoses/blob/3d4944359cddb9d66bd0995b6175a22515de08fb/app/models/network/technologies/deferrable_consumer.rb#L117). It was designed to "dump" any remaining postponed loads onto the network in the final frame of the year (since that's the last opportunity to satisfy that demand).

**Comment by ChaelKruip** [comment]
> It was designed to "dump" any remaining postponed loads onto the network in the final frame of the year

That sounds right to me. It seems to me though that @grdw observes the following behaviour: **all** load that has ever been postponed is 'dumped' in the last frame. This would be incorrect IMO as the maximum duration of postponing is 3 hours.

@grdw can you confirm that this is what you observe? Or am I muddling a perfectly clear conversation?

**Comment by grdw** [comment]
> @grdw can you confirm that this is what you observe? 

The spec is the only thing that produces this behavior.

screen shot 2016-03-22 at 14 00 22

The development environment does give the correct result it's a bit weird.

**Comment by ChaelKruip** [comment]
> The development environment does give the correct result; it's a bit weird.

That image looks about right indeed. So, does your spec allow more than 12 steps between the moment of postponement and the end of the curve?

**Comment by ChaelKruip** [comment]
After discussion with @grdw it became clear that the current spec needs some changes in order to produce realistic results. The necessary changes are underway.

**Comment by antw** [comment]
[There is a very low-level test already](https://github.com/quintel/etmoses/blob/3d4944359cddb9d66bd0995b6175a22515de08fb/spec/models/network/technologies/deferrable_consumer_spec.rb#L48-L51), but that checks the technology in isolation, not as part of a network. Still, it should be enough to prove that loads cannot be deferred longer than 12 frames (3 hours) before becoming mandatory.

**Comment by grdw** [comment]
> Still, it should be enough to prove that loads cannot be deferred longer than 12 frames (3 hours) before becoming mandatory.

Yes. I agree. It becomes a little bit of a time-waste to do that.

However could you than maybe elaborate on:

> I'm especially curious if the calculation of the ratio (for base loads) is still correct when only using a week of data. I suspect not, since profiles used for "Postponing of individual base load" do not have a fixed ratio whereas the EDSN profiles do..

What is it we should be trying to test?

**Comment by antw** [comment]
> What is it we should be trying to test?

For base loads there are flexible and inflexible components. With EDSN profiles, the ratio between the flexible and inflexible is constant throughout the year. [Note how the two curves have an identical shape](https://beta-moses.energytransitionmodel.com/load_profiles/240); one is just bigger than the other. It might be that the flexible component accounts for 30% of the demand, and the inflexible is 70%. Since the ratio is constant, we could sample any week and the ratio will always be 30%/70%.

However, with individual profiles, this is not the case. [The flexible and inflexible components have a different shape](https://beta-moses.energytransitionmodel.com/load_profiles/311). The ratio between the two curves for the entire year might be 30%/70%, but if you sample only the first week, there may be more inflexible loads and the ratio is instead 40%/60%. Since this ratio is used [to scale the curves](https://github.com/quintel/etmoses/blob/3d4944359cddb9d66bd0995b6175a22515de08fb/app/models/installed_technology.rb#L285), this *might* result in the technology having the wrong loads.

More concretely: in the weekly calculation, does a technology using an individual base load have an identical load in the first week of the year as it currently does on beta? If so, we have nothing to worry about. :smiley:

**Comment by grdw** [comment]
> More concretely: in the weekly calculation, does a technology using an individual base load have an identical load in the first week of the year as it currently does on beta? 

We can - to be sure - always calculate the ratio for the whole year instead of for a ratio?

**Comment by antw** [comment]
> We can always calculate the ratio for the whole year

:+1:

**Comment by grdw** [comment]
I managed to clean up the curve length a little bit. I did solve it a bit different but I'd like to know your opinion.

**Comment by antw** [comment]
> I managed to clean up the curve length a little bit. I did solve it a bit different but I'd like to know your opinion.

I like extracting the profile attributes to a new class, that's much better design. But I'm a little confused as to the distinction between an `InstalledTechnology::ProfileCurve` and `CurveCollection`. It seems like both classes do much the same thing: wrap around one or more load profiles. I wonder if these two classes would be better if merged together?

(Since `CurveCollection` is only used for load profiles, it could be renamed to `LoadProfileCollection`; or `InstalledTechnology::LoadProfileCollection` if you wanted to namespace it).

--

The main remaining issue I note is that profiles are incorrectly scaled when a technology has a `demand` attribute. I think the only solution will be to read and scale the full curve, then extract the data for the week we care about (dividing demand by 52 won't suffice, since demand is unlikely to be evenly distributed throughout the year).

This affects all technologies which have a `demand` attribute including base loads and buffers. Here's an example using an individual base load:

```
Endpoint:
  capacity: 50.0
  technologies:
    1x Household (postponing base load)
       profile: anonimous_base_load_1
       demand: 8760.0
```

| master branch | pull request |
|---|---|
| master | pr |

**Comment by ChaelKruip** [comment]
> The main remaining issue I note is that profiles are incorrectly scaled when a technology has a demand attribute. 

Well spottend @antw! It seems that the week is scaled using the annual demand indeed. 

> I think the only solution will be to read and scale the full curve, then extract the data for the week 

:+1:

**Comment by grdw** [comment]
@antw if I understand correctly, we should go from:

````ruby
def scaled_network_curve(scaling, range)
  Network::Curve.new(
    case scaling
    when :capacity_scaled then Paperclip::ScaledCurve.scale(network_curve(range), :max)
    when :demand_scaled   then Paperclip::ScaledCurve.scale(network_curve(range), :sum)       
    else network_curve(range)
    end.to_a
  )
end
````

to:

````ruby
def scaled_network_curve(scaling, range)
  range = range.nil? ? 0..35040 : range

  Network::Curve.new(
    case scaling
    when :capacity_scaled then Paperclip::ScaledCurve.scale(network_curve, :max)
    when :demand_scaled   then Paperclip::ScaledCurve.scale(network_curve, :sum)
    else network_curve
    end.to_a[range]
  )
end
````

**Comment by antw** [comment]
> if I understand correctly, we should go from:

Something like that. However I also think we should avoid passing around `range` attributes into half-a-dozen different methods; it's beginning to get messy. Perhaps it can be wrapped up into `InstalledTechnology#get_profile` or a similar method, and kept in one place?

I also noticed new `InstalledTechnology#profile_range` `#preset_load_profile` and `#curve` attributes. Like `profile_length`, these aren't attributes of the technology, but "helpers" used to set up a calculation. These could also be wrapped up into a separate class (perhaps the new ProfileCurve class) responsible for setting up the profiles (including extracting the date-range needed), ready for calculation. The `InstalledTechnology` class is getting very large and bloated. :anguished:

**Comment by grdw** [comment]
> The InstalledTechnology class is getting very large and bloated. 

Yes. I noticed. I'm busy cleaning it out.

**Comment by antw** [comment]
Maybe I can be clearer about what I mean. The InstalledTechnology class was meant to store attributes about a technology, chosen by a visitor. It already violated the [single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle) by also having to scale profiles, deal with ratios, etc. But now it also has to extract a date range from those profiles, this means passing a `range` attribute into several different methods. All those other methods are made more complicated as a result.

When this happens, the nice solution is often to add a new layer *on top* of the old functionality: for example, a new class whose sole responsibility is to take the profile curves from the technology, and extract the date-range we want:

```ruby
def profile_curve(range)
  RangeExtractor.new(get_profile, range)
end
```

In a perfect world, we might have several of these classes to get the data into the format we want, instead of burdening the inner class (the technology) with these responsibilities.

grdw:
I managed to move the range attributes out of a lot of classes/modules. InstalledTechnology still looks pretty crappy though. I also don't get why we shouldn't just make it easier and remove the ability to use a plain Array as a profile (only allow static Hashes in test cases) - in order to make `get_profile` a bit more readable (with 16 lines it's a little bit long).

grdw:
Allright, I changed some more! :dancers:  I managed to do it a little as you described without the extra classes yet. However if you could have a look, that would be great! :smile:

antw:
Looks much improved! I'll give it a test shortly, and I think we can expect to merge this. :+1:

antw:
Merged; thanks @grdw! I think our visitors will really appreciate the faster response times from weekly calculations. :tada:

PR #872, opened 2016-03-25
Issues/#800 showing parts of load

grdw:
This is basically a merge of PR #837  + @antw loads per tech + my experimental stacked view.

Things todo:

- [x] Incorporate loads per tech data in the stacked chart
- [ ] Loads of testing
- [ ] ... Fix what comes out of the above point


grdw:
Some previews:
screen shot 2016-03-25 at 14 34 32
screen shot 2016-03-25 at 14 34 53

- Switching between gas and electricity is working
- You can switch from 'Total', 'Stacked' and 'Invdividual' 
- Strategies is something which needs to be tested (probably doesn't work yet) :sweat_smile:

grdw:
Some more previews (strategies on/off):
screen shot 2016-03-25 at 17 05 11
screen shot 2016-03-25 at 17 05 19

dennisschoenmakers:
:+1: Looking forward to try this out on some LESS's! :smiley:

ChaelKruip:
Nice work @grdw!

I have some suggestions for improvement:

* The legend can be structured better: the coloured square should be on the same line as the name of the technology it represents. Perhaps we can hide all technologies that have zero load to make the legend less cluttered?
* The colours of technologies curves should not change with toggling between strategies 'on' and 'off' because that will confuse users.

grdw:
> The legend can be structured better: the coloured square should be on the same line as the name of the technology it represents. Perhaps we can hide all technologies that have zero load to make the legend less cluttered?

Ehm like this:

````
[ ] Test
[ ] Technology 1
````

Or spaced out like
````
[ ] Test         [ ] Technology 1
[ ] Technology 2 [ ] Test 2
````


> The colours of technologies curves should not change with toggling between strategies 'on' and 'off' because that will confuse users.

Hmm.. agree. I did it to see if there actually changed anything. :smile:

**Comment by ChaelKruip** [comment]
> Or spaced out like

I like this better because it will use the available space better.

**Comment by grdw** [comment]
screen shot 2016-03-30 at 17 18 10

**Comment by dennisschoenmakers** [comment]
Great work @grdw! My 2 cents: could you try to turn the colored square and the label around? Just like in etmodel:

image

And I think the hover over part with the numbers per series can be improved a bit. Now it kinda floats there: perhaps by giving it a (highly transparent?) backgroundcolor and a border?

**Comment by grdw** [comment]
I cleaned up the gitlog because it was a horrible mess. 700 merge conflicts later: tada! :dancers:

**Comment by dennisschoenmakers** [comment]
Exciting!

**Comment by antw** [comment]
The charts look good. The [triangular EV load here](https://github.com/quintel/etmoses/pull/872#issuecomment-201345644) suggests I may still have some tweaking to do; I'll look into it and do some testing.

I'd like the weekly loads to be on beta for a few days before we merge this, so that people get a chance to test it and verify that everything is working as expected. That way when we do merge this one, if there are strange numbers, we know to blame my calculator changes and not the weekly calculation. :laughing:

**Comment by grdw** [comment]
I do spot some odd results in this overview.. 

Could be that it is actually correct but it seems highly unlikely.

This is the MV trafo (directly above the second chart):

screen shot 2016-04-01 at 10 35 19

This is the LV1 trafo (directly under the first chart):

screen shot 2016-04-01 at 10 35 26

What I see is that EV has taken up alot of load. Either this is correct or something doesn't really add up (also the shapes of the charts are different - total load and stacked load). Another thing I noticed is that my Solar PV y-values are returned as an array with zero's (that is before stacking them). 

screen shot 2016-04-01 at 10 48 00

**Comment by grdw** [comment]
> Great work @grdw! My 2 cents: could you try to turn the colored square and the label around? Just like in etmodel:

@dennisschoenmakers 

The reason I aligned it to the right is purely estethic. When I align it to the left there's this white bar of nothing next to the chart which is really disturbing to the eye (see picture below). The reason that there's that much space at the left side of the chart is because of some LES's running their numbers into the 100.000's (like Lochem). 

screen shot 2016-04-01 at 10 55 53

**Comment by dennisschoenmakers** [comment]
> When I align it to the left there's this white bar of nothing next to the chart which is really disturbing to the eye (see picture below).

I don't find it that disturbing. Of course, it would be even nicer if we could left align the chart depending on the labels, but this is already good enough IMO.

**Comment by grdw** [comment]
I'm doing some sanity checking by comparing to production at the moment to see if the loads are somewhat the same. 

| Technology                          | Ok?  |
|------------------------------------- | -------- |
| EV                                        | See comment above |
| Solar Panel                          | See comment above |
| Buildings                              | All is fine |
| Households                          | All is fine |
| HP (Hot water air)                | All is fine |
| HP (Hot water ground)         | All is fine |
| HP (Space heating ground)  | All is fine |
| HP (Space heating air)         | All is fine |

Power-to-Gas is something of a special case I guess. But there's something with negative loads which results in zero's.. 

Will be expanded

File: app/models/installed_technology.rb [hunk 1]
```diff
@@ -276,8 +276,10 @@ def valid_profile?
     end
   end
 
-  def is_battery?
-    %w(congestion_battery households_flexibility_p2p_electricity).include?(type)
+  def no_profile?
+    %w(congestion_battery
```
antw: If this must be hard-coded, define it in a constant somewhere so that a new array and three strings aren't created every time `no_profile?` is called. However, this seems like it would be best as a boolean column in the `technologies` table (something like `requires_profile default=true null=false ...`).

My general rule is that if I ever find yourself hard-coding the name of something from the database, especially unique keys such as these, try to find a way to move the information into the database and out of the code. Otherwise the application becomes very fragile to changes; if ever the names of these technologies were to change, `no_profile?` would return the wrong value.

antw:
I'm fairly sure I know what causes the anomalous EV loads, and if so it will also be affecting the P2P battery. It's down to an implementation detail which allows those techs to release enemy into the network to be used by other technologies, but is reclaimed by the battery if is not used.

Unfortunately a solution eludes me at the moment. I'll keep thinking. :worried:

Issue #881, opened 2016-03-31
Some weeks do not show loads with weekly calculations


ChaelKruip:
[This LES] shows no loads for the second week of January:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/14179516/a7077a02-f75f-11e5-9db7-e7206f85b209.png)

grdw:
Hmm,, I see.. crap. I'm looking into it as we speak.

grdw:
Well I figured out it's a caching problem.. (ofcourse). It turns out it cuts out a piece of 673 array items at the range of 673..1346 . which ofcourse is not possible. It should only do that when it has a full year in cache.

PR #882, opened 2016-03-31
Fix loads per week issues for electricity load

grdw:
Fixes: #881 

To reproduce: 
1. turn on caching. 
2. Add some gas technologies. 
3. Change the week 
4. Cry

Unfortunately this doesn't yet work for gas for some yet to be discovered reason... It now just shows the load of week 1 - always for gas. It annoys me but I can't figure where it fetches the loads of week 1 and for gas technologies only. I'll look into it tomorrow.

### DIFF
--- a/app/assets/javascripts/les/topology_load_charts/chart_shower.js
+++ b/app/assets/javascripts/les/topology_load_charts/chart_shower.js
@@ -109,7 +109,7 @@ var ChartShower = (function () {
                 return false;
             }
 
-            window.currentTree.nodes.push(this.nodeData.name);
+            window.currentTree.addNode(this.nodeData.name);
             renderLoadChart.call(this, this.nodeData);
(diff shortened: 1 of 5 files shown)


grdw:
@antw could you explain to me why. [Here](https://github.com/quintel/etmoses/blob/master/app/models/installed_technology/profile_curve.rb#L37) I have the correct cut out of gas load curve and [here](https://github.com/quintel/etmoses/blob/master/app/models/calculation/technology_load.rb#L38) curve is nil? I'm trying to wrap my head around it (I turned off caching but gas is still only showing the first week of the load eventhough it correctly cuts out the week... ). But each should yield the correct curve instead it yields nil. Is it because technologies attached to a composite should not have a profile?

grdw:
Thansk :+1:

Issue #874, opened 2016-03-30
Concurrency buttons can be drastically reduced


ChaelKruip:
The buttons with a line through it can be removed:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/14136778/e3305040-f664-11e5-836d-d6af9b31b230.png)

grdw:
![screen shot 2016-03-31 at 11 35 00](https://cloud.githubusercontent.com/assets/2676542/14171706/9e23c376-f734-11e5-858e-840f8184c84e.png)

That will leave us with this :+1: 
There are boolean's called 'expandable' in the `technologies` table. I set them manually on the beta environment to false for the technologies specified in the image above.

ChaelKruip:
:+1: nice changes @grdw!

One last request: can you make all names plural? "Households" instead of "Household" etc.

grdw:
> One last request: can you make all names plural?

Yes. I could. I do consider this somewhat low-prio so I'll probably change this tomorrow.

ChaelKruip:
> Yes. 

:+1: 

> I could. 

:+1: 

> I do consider this somewhat low-prio so I'll probably change this tomorrow.

:+1:  :+1:

grdw:
<img src="https://media.giphy.com/media/XreQmk7ETCak0/giphy.gif"/>

PR #883, opened 2016-04-01
Issues/concurrency settings

grdw:
Fixes: #874

### DIFF
--- a/app/views/testing_grounds/_technologies_form.html.haml
+++ b/app/views/testing_grounds/_technologies_form.html.haml
@@ -32,7 +32,7 @@
       %span.check_box
         - max = maximum_concurrency?(technology.key, testing_ground.technology_profile)
         = check_box_tag technology.key, "max", max, {data: {toggle: "toggle", on: "Max", off: "Min"}}
-        %span= technology.name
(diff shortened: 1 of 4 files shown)


app/views/testing_grounds/_technologies_form.html.haml
```diff
@@ -32,7 +32,11 @@
       %span.check_box
         - max = maximum_concurrency?(technology.key, testing_ground.technology_profile)
         = check_box_tag technology.key, "max", max, {data: {toggle: "toggle", on: "Max", off: "Min"}}
-        %span= technology.name
+        %span
+          - if technology.composite?
+            = technology.name
+          - else
+            = technology.name.pluralize
```
antw: This would be ideally done using the locale YAML files, such as:

```yaml
technologies:
  transport_car_using_electricity:
    singular: Electric car
    plural: Electric cars
  buffer_space_heating:
    singular: Buffer space heating
    plural: Buffer space heating
```

```haml
%span= t("technologies.#{ technology.key }.plural")
```

grdw: Good point. I'll change this :+1:

PR #887, opened 2016-04-05
Issues/#878 Adding pressure level sheet (Part 1/3)

grdw:
This is the initial gas asset list feature.

The data used for the pipes and connectors in this is pull request is as fake as it gets. 

However I do have some questions before merging:

- The way I initialize the static data, is that 'okay'? (See `config/initializers/active_hash.rb`)
- Also I have some doubts about the data-formatting of the fake data. But I'll create an issue about it (unless you already have some ideas about it @antw ?)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,9 @@ gem 'delayed_job_active_record'
 # Caching
 gem 'msgpack'
 
+# Active hash
+gem 'active_hash'
+
 group :development, :test do
   gem 'rspec-rails', '~> 3.1'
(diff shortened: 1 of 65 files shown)


antw:
> The way I initialize the static data, is that 'okay'? (See config/initializers/active_hash.rb)

I think it's fine. I like that the data is converted into proper objects (`Pipe` and `Connector`) instead of just being raw hashes. :+1:

I'm using this PR as the basis for combining the gas network front-end with the back-end calculation, so I'll merge both together when ready.

Issue #756, opened 2016-02-11
Technology template  contains old default values when trying to add a new technology.


grdw:
In the profile matrix there's a HTML template we update when changing the technologies select box. However there are cases when that HTML template contains values from a previously selected technology. 

This is fixable by removing it and adding it again.

ChaelKruip:
Putting this on-hold as it is not urgent

PR #899, opened 2016-04-13
Fixes incorrect attributes for technologies when adding them to the profile matrix

grdw:
Fixes: #756


antw:
Are you happy that this works as expected? If so, I think it's fine to push simple changes like this straight to master without needing a review. Unless you *want* a second opinion of course, in which case you can always create a PR.

(I did test it, seems to work just fine :+1:)

grdw:
Great! Well, I just create a PR to be sure that somebody has also seen it (i.e. is aware that something changed eventhough it's something little). It's not really about the second opinion (unless you have a really strong one). 

I could from now on just merge small things like this directly into master. Can imagine that it might be disruptive of your workflow to look at something this small.

PR #909, opened 2016-04-15
Issues/#878 adding fake stacked gas asset data (Part 2/3)

grdw:
Adding a gas load page with a gas chart (the data is used is fake).

Merge: #887 first.


antw:
@ChaelKruip This is what it looks like with some real data.

screen shot 2016-04-18 at 14 30 44

The `0.125 bar ↔ 4 bar` and `4 bar ↔ 8 bar` connections have some fairly extreme constraints, and 50% loss.

On the one hand, I think this chart is useful for getting an overview of the loads on each layer, but I also find the combination of capacity constraints and loss can be quite confusing. A visitor might not understand how there could be 14MWh of load on the endpoints, but only ~1MWh the next level up. The answer is that there's a capacity constraint, and a 13MWh deficit.

We can remedy this by showing deficits and surpluses (flaring) at each level:

screen shot 2016-04-18 at 14 29 43

What do you think?

Loss is a bit of an awkward one, because it is visualised on the positive (consumption) side of the chart, when in fact it can occur both as a result of consumption *and* production (feed-in). Perhaps it should be split into two separate parts?

ChaelKruip:
> We can remedy this by showing deficits and surpluses (flaring) at each level:

I like this a lot! I suggest to plot the deficit in the positive direction (just as consumption). 

The negative direction is reserved for feed-in and I think both the feed-in and the resulting loss should be plotted in the negative direction.
I would rename "Surplus" to "Flaring" and show it in the negative direction as well. 

FYI: the losses only occur for feed-in direction (when compressing gas from low to high pressure). Gas flowing down has no losses.

antw:
> I would rename "Surplus" to "Flaring" and show it in the negative direction as well.

Will do.

> FYI: the losses only occur for feed-in direction (when compressing gas from low to high pressure). Gas flowing down has no losses.

I know that's true in the real-world, I just wanted to cover all eventualities. 😆 

I'll swap loss to the negative side of the chart and leave it at that for now.

PR #918, opened 2016-04-18
Fix load profiles preview chart page

grdw:
Just saw that the load charts page is not showing any charts. This PR fixes that problem.


grdw:
One moment please.

PR #922, opened 2016-04-19
Expanding business case calculation with gas asset list

grdw:
I created a separate pull request for the business case expansion, instead of the previous one (that branch was too messed up to fix :sweat_smile: ).


app/models/network/node.rb
```diff
@@ -199,6 +199,11 @@ def units
       get(:units) || 1
     end
 
+    # Public: A string that returns the stakeholder
+    def stakeholder
```
antw: This method is the only thing I don't like about this PR – and it's a very minor thing, I know. 😆 

The stakeholder isn't an attribute relevant to the network calculation – i.e., the calculation of loads – so I don't think it belongs here. It's not guaranteed that this attribute will always be available (for example, if the `Network` module is extracted out to a separate gem), so I'm reluctant to add it as a first-class method on `Node`.

Can I suggest just using `get(:stakeholder)` for now, until I decide what I'm going to do with the `Network` module in the longer-term?

grdw: > Can I suggest just using get(:stakeholder) for now, until I decide what I'm going to do with the Network module in the longer-term?

Yes, that's fine for me :+1: I'll see what I can manage.

grdw: I removed the stakeholder method and solved it somewhat different.

PR #923, opened 2016-04-19
Issues/gas asset list charts for loads

grdw:


antw:
I cherry-picked onto master the stuff which seemed relevant.

I did notice two minor issues:

The "hover line" values often don't appear on the gas load chart. Sometimes they do, but in the wrong place (too high, above the chart):

screen shot 2016-04-19 at 18 28 16

The "40 bar" bar is chopped off (probably just a chart width issue?)

screen shot 2016-04-19 at 18 28 21

grdw:
> Sometimes they do, but in the wrong place (too high, above the chart):

I'll check to see what's the problem.

> The "40 bar" bar is chopped off (probably just a chart width issue?)

Yes it is. I'll take a look.

grdw:
@antw one question. Shall I remove the business case commit from this PR?

antw:
Thank you!

I had to bypass the cache (temporarily) for the gas charts, so I especially appreciate the spinners while they load. Sometimes it's the little things that matter. 😀

PR #928, opened 2016-04-19
Renaming of static pipes + connectors

grdw:
Renaming pipes to something a bit more serious.

For presentation purposes.


antw:
@ChaelKruip FYI: This might break scenarios you have already been working on, on account of the new names for the connectors. Is that a problem? If so, we can remedy this by adding a migration to update existing LESes.

ChaelKruip:
> Is that a problem? 

No problemo!

PR #942, opened 2016-05-02
Associating importable attributes on technology keys instead of ids

grdw:


db/migrate/20160502145559_remove_technology_id_field_from_importable_attributes.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveTechnologyIdFieldFromImportableAttributes < ActiveRecord::Migration
+  def change
+    remove_index :importable_attributes, :technology_id_and_name
+    remove_column :importable_attributes, :technology_id
+    add_index :importable_attributes, [:technology_key, :name], unique: true
```
antw: This would be a good opportunity to add a validation to ImportableAttribute to catch duplicate indices before attempting to save.

```ruby
validates :name, uniqueness: { scope: :technology_key, case_sensitive: false }
```

grdw: Yes and no. Yes, because it's a good idea. But no because I'm currently trying to move all technologies to yaml files and in that case I'm getting rid of the `importable_attributes` table and therefor this file.

antw: Makes sense, thanks. :+1:

antw:
If this is part of the move to YAML-based technologies, would it not be better to do both at once? It seems like this work will just end up being deleted once the YAML techs are implemented.

grdw:
> If this is part of the move to YAML-based technologies, would it not be better to do both at once?

Yes. That's better 👍  I'll close this PR.

Issue #947, opened 2016-05-03
LES crashes while applying different flex strategies or selecting other nodes


Sanne512:
When applying different flex strategies or selecting other nodes in a LES on the beta server the LES crashes. The specific LES is on @JeroenvdLogt 's account.

PR #949, opened 2016-05-03
Fix selecting nodes which have no gas load

antw:
@grdw Could you check that this change makes sense? When caching was enabled, nodes which have no gas would raise an error in `isValidNodeData` because `d.gas.total` would be `undefined`. This change checks that the `total` is set before asserting the array length is non-zero.

I tested locally and the charts for all nodes seem to load correctly.

Ref #947.


grdw:
This makes sense! 👍 I'll merge it.

PR #950, opened 2016-05-03
Issues/#740 export technologies to yaml files

grdw:
Before testing:

# ! Please make an export of your local database. !

Than proceed to run the migrations and run the code.


db/migrate/20160502145133_rename_technology_ids_to_keys.rb
```diff
@@ -0,0 +1,10 @@
+class RenameTechnologyIdsToKeys < ActiveRecord::Migration
+  def change
+    add_column :importable_attributes, :technology_key, :string, after: :technology_id
+
+    ImportableAttribute.all.each do |ia|
```
antw:
```
uninitialized constant RenameTechnologyIdsToKeys::ImportableAttribute
moses/db/migrate/20160502145133_rename_technology_ids_to_keys.rb:5:in `change'
NameError: uninitialized constant RenameTechnologyIdsToKeys::ImportableAttribute
```

`ImportableAttribute` class has been removed so this migration won't work anymore. I'm guessing it can just be removed?

grdw: Oh, yes it can.

antw:
I think this is good, and I'm glad we're moving this data out of the database. A couple of comments:

* Instead of `FILE_ROOT` being hard-coded in `Technology`, how about using [`Settings.static_data_path`](https://github.com/quintel/etmoses/blob/a271d5e494ace7a03ee51bf8b133675c30c8b31c/config/settings.yml#L6-L7) so that the tests can use their own static files? This is quite useful because it means that tests don't become dependant on real-world data; if a tests checks that a demand of *x* and a capacity of *y* results in load *z*, changing a tech capacity might cause the test to fail. In fact, the code is working just fine (tests are for verifying code, not data).

  For example, both [Atlas](https://github.com/quintel/atlas/tree/master/spec/fixtures) and [ETEngine](https://github.com/quintel/etengine/tree/5e2f6fe33eff64fd4dc0e4e0d1330ca4eb33ac07/spec/fixtures/etsource) have their own stripped-down versions of ETSource in their tests. It would be pretty awful if every update of ETSource required us to update tests in those two projects; having their own versions of ETS keeps the tests reliable and repeatable.

  The way to go about this might be to add the ability to use the `Settings` path – in the test environment its value is `spec/fixtures/static` – and, with *no* technology YAML files in that directory, run the tests with the `--fail-fast` option. If nothing fails, great! Something likely *will* fail, so copy the technology file necessary to get the tests to pass and then try it again. The aim is to have the minimum number of technology.yml files required to get the tests to pass.

* **Nice-to-have**: It would be nice to be able to omit some attributes from the technology files and have sensible defaults assumed. For example, we could remove any attribute which has no value set (plenty have `export_to:` and `default_capacity:`).

  A default could be set for attributes like `visible`, `expandable`, and `composite`, with the attribute only appearing in the YAML file if it differs from the default.

* **Nitpicking nice-to-have**: Perhaps instead of `default_capacity`, `default_volume`, and `default_demand`, we might have a `defaults` hash instead?

  ```yml
  defaults:
    capacity: 4.0
    volume: 10.0
  ```

* **Nitpicking nice-to-have**: It would also be nice to remove the `name` attribute from the YAML files, and have this defined in a locale file.

grdw:
> Instead of FILE_ROOT being hard-coded in Technology, how about using Settings.static_data_path so that the tests can use their own static files

Done! 

I also removed all the blank keys. I also once looked at those defaults (for expandable, visible etc.) but there's something odd going on there which I couldn't explain directly. I'll try it again though to see what that was. 

I also agree with the defaults Hash but I'll do that some other time.

antw:
Thank you @grdw! This is nice.

I removed the migration on `ImportableAttribute` and created an extra backup of the staging DB as suggested. To `master` it goes!

PR #953, opened 2016-05-09
Issues/separate business case

grdw:


app/assets/javascripts/les/topology_load_charts/gas_load_chart.js
```diff
@@ -39,6 +39,8 @@ var GasLoadChart = (function () {
 
     GasLoadChart.prototype = {
         draw: function () {
+            $("ul.nav.nav-tabs li.gas_load").addClass("active");
```
antw: I'd like to see the menu stuff moved out of `GasLoadChart`, `Tree`, etc. Activating a menu item isn't their responsibility, and would make it harder to reuse those features in the future (e.g. maybe you might want to put a gas chart on the business case page for some reason...).

Also, don't forget to be specific about *which* menu you're working with: `#menu ul.nav`, and not just `ul.nav` (of which there may be more than one).

You might set up a single event handler to do this:

```js
$(document).on('page:change', function () {
  var lesNav = $('#menu ul.nav'), activeEl;

  if (window.location.pathname.match(/gas_load/)) {
    activeEl = lesNav.find('.gas_load');
  } else if (...) {
    // ...
  }

  lesNav.find('li').removeClass('active');

  if (activeEl) {
    activeEl.setClass('active');
  }
}
```

Better yet would be a proper class:

```js
var lesMenu = new LesMenu();

$(document).on('page:change', function () {
  lesMenu.activateWithPath(window.location.pathname);
});

```

grdw: Done! See https://github.com/quintel/etmoses/blob/issues/separate-business-case/app/assets/javascripts/les/les_menu.js

PR #958, opened 2016-05-11
Issues/gas asset yaml updates

grdw:


config/initializers/active_hash.rb
```diff
@@ -1,10 +1,8 @@
 Rails.application.config.to_prepare do
-  unless defined?(DATA_SOURCES)
-    DATA_SOURCES = {
-      "connectors" => GasAssets::Connector,
-      "pipes"      => GasAssets::Pipe,
-    }.freeze
-  end
+  DATA_SOURCES ||= {
```
grdw: @antw I had some issues with the `unless defined`. Mainly because my `GasAssets` would be empty from time to time on development but with this code it isn't.

antw: Okay! I added that because Ruby keeps complaining about redefining the constant (I'm guessing when Rails reloads code), but it wouldn't surprise me if the reloader is also why your GasAssets were empty.

I believe the reloader works by removing the old class and creates a new one. I'm *guessing* that `DATA_SOURCES` will still be referencing the *old* classes. I also suspect this is the culprit behind quintel/etengine#809.

We might be able to get around that by storing the class name as a string, instead of a direct reference to the class:

```ruby
unless defined?(DATA_SOURCES)
  DATA_SOURCES = {
    "connectors" => "GasAssets::Connector",
    "pipes"      => "GasAssets::Pipe"
  }.freeze
end

DATA_SOURCES.each_pair do |folder, class_name|
  class_name.constantize.data = # ...
end
```

grdw: >  Okay! I added that because Ruby keeps complaining about redefining the constant

Yes, I remember those messages. With this line I also don't see those warnings anymore though. It mostly happened when I had multiple rails consoles open. But right now I can't seem to reproduce it... (when switching back to the old code aswell). Also running `Rails.cache.clear` from the console doesn't seem to have any effect.

Hmm it might be. I'll give it a shot.

grdw:
Crap, something slipped in this PR which shouldn't be there. Moment.

PR #965, opened 2016-05-13
Issues/adding compressors to gas asset yaml

grdw:
This is not battle tested yet but using the original specs I'm able to set the `low_to_high` capacities and efficiences with the help of compressors. 

- [x]  Test it in a real LES


grdw:
It doesn't freeze a LES except for reasons specified here https://github.com/quintel/etmoses/issues/996 .

grdw:
@antw ,, hmm I might accidentally removed some commits of yours, could you maybe check if that's the case?

antw:
>  hmm I might accidentally removed some commits of yours, could you maybe check if that's the case?

I don't see anything my old local copy of this branch. It might have been [this merge commit](https://github.com/quintel/etmoses/commit/e3b08fea21eb355c487d90758bfc2511b7f3fa60), but that's no problem as it's already on `master`.

grdw:
> I don't see anything my old local copy of this branch.

Than all is fine I guess 👯  was just checking to be sure

antw:
Merged manually, thanks!

Issue #805, opened 2016-02-18
Losses should be taken into account for all technologies


ChaelKruip:
As discussed in https://github.com/quintel/etmoses/issues/799, losses are not currently taken into account for technologies except for heat pumps.

## Heat pumps

Currently, for heat pumps we distinguish two types of capacity:
* `capacity`: the amount of heat the pump can deliver per unit time in kW
* `carrier_capacity`: the amount of carrier (in this case electricity) that the technology consumes per unit time in kW

The two capacities are related by the `performance_coefficient` which is equal to the `coefficient_of_performance` (COP) as defined in ETEngine by:
```
capacity =  carrier_capacity * performance_coefficient
```
As the `performance_coefficient` is larger than unity, the `capacity` is larger than the `carrier_capacity`.

This behaviour is in line with the ETM and no further actions are needed for heat pumps

## Heaters, hot water producing technologies and the gas-part of the hybrid heat pump

These technologies are currently only described by one `capacity` which determines both the input (gas, electricity) and the output (heat, hot water) of the technology. Losses are not taken into account.

I think it would make sense to describe such technologies by an `output_capacity` and an `efficiency`. The load on the network (gas or electricity) would then be given by an `input_capacity` which follows from the previous two by:
```
input_capacity = output_capacity / efficiency.
```
The `input_capacity` would typically be larger then the `output_capacity` because `efficiency` is smaller then unity in general. One notable exception is the `households_space_heater_combined_network_gas` (and the gas-part of the hybrid heat pump for space heating which is basically a  `households_space_heater_combined_network_gas`):
```
+----------------------------------------------------------+------------------------+-----------------------+----------------------+
|                           key                            |        efficiency      | input_capacity [MW]   | output_capacity [MW] |
+----------------------------------------------------------+------------------------+-----------------------+----------------------+
| households_water_heater_network_gas                      | 0.67                   | 0.035820895522388055  | 0.024                |
| households_water_heater_combined_network_gas             | 0.9                    | 0.024444444444444442  | 0.022                |
| households_water_heater_resistive_electricity            | 0.95                   | 0.0015789473684210528 | 0.0015               |
| households_space_heater_electricity                      | 1.0                    | 0.006                 | 0.006                |
| households_space_heater_network_gas                      | 0.8                    | 0.027499999999999997  | 0.022                |
| households_space_heater_combined_network_gas             | 1.067                  | 0.020618556701030927  | 0.022                |
[...]

PR #1019, opened 2016-05-31
Adding basic heat source list

grdw:
First merge all et-engine related changes:

https://github.com/quintel/etengine/pulls


antw:
TY! I'm going to get the heat network calculator rebased on top of this, and will get them connected up. I'll merge both together at the same time (with the ETE pull requests also).

antw:
Are there any commits still to be pushed here or for ETEngine? I'm finding that Moses is asking for attributes which either haven't been whitelisted in ETE, or are private methods on ETEngine's ConverterApi (and therefore fail).

grdw:
Yes, I forgot to push them. I now did.

app/models/installed_heat_source.rb
```diff
@@ -0,0 +1,18 @@
+class InstalledHeatSource
+  include Virtus.model
+
+  attribute :key, String
+  attribute :name, String
+  attribute :units, Float
+  attribute :heat_production, Float
```
antw: I'm curious about the names `heat_production` and `heat_capacity`. Are these the heat technology versions of `demand` and `capacity`? If so, what motivated the difference in naming? Is there a limitation in the ETEngine importer such that we have to use different names?

grdw: > I'm curious about the names heat_production and heat_capacity. Are these the heat technology versions of demand and capacity?

You could say so. I think the prefix `heat_` is a bit redundant which I think is what you are trying to tell me (?). I am a bit freightened about the `capacity` term since there's an `input_`- and `output_capacity` and @ChaelKruip has a strong opinion about this matter.

> Is there a limitation in the ETEngine importer such that we have to use different names?

Nope

antw: > I am a bit freightened about the capacity term since there's an input_- and output_capacity and @ChaelKruip has a strong opinion about this matter.

Okay! The capacity attribute is not a huge deal; I can work around that and get the heat producers to call `heat_capacity` instead of `capacity`.

Regard `production` though, wouldn't calling it `demand` make it easier to get a demand-scaled network curve? I know that's an InstalledTechnology method, but I'll need to add that to InstalledHeatSource too; maybe that's not a big deal either, since the heat-source version doesn't need to be as complex as the technology one...

ChaelKruip: > I am a bit freightened about the capacity term since there's an input_- and output_capacity and @ChaelKruip has a strong opinion about this matter.

Me having a strong opinion about something should not be a motivation to choose a sub-optimal solution here. Do whatever makes more sense code-wise. 

To clarify: my opinion is about the fact that technologies in general have a *set* of capacities:
* output_capacity: the rate at which energy can be deliverd by the technology
* input_capacity: the rate at which energy can be fed into the technology

Without having consistent and physically meaningful names for such basic variables, it will be very difficult to prevent/detect errors made in the energetic calculations. Moreover, adding functionality as described in https://github.com/quintel/etmoses/issues/805 will be a pain.

grdw: > Regard production though, wouldn't calling it demand make it easier to get a demand-scaled network curve?

How do you mean? This might seem like a stupid question: but Is the `demand` of a heat technology equal to the `production`?

antw: > This might seem like a stupid question: but Is the demand of a heat technology equal to the production?

Yes and no. Demand describes the total amount of energy in kWh a technology will consume in a year, production is the total amount if will produce. In the case of heat sources, I was thinking about it the same way we do with capacity, where a positive number shows consumption and a negative indicates production. I was expecting the heat sources to have a negative demand.

Maybe this is not so important: a positive "production" will certainly make more sense to a user, so perhaps we should keep it the way it is.

We do still need a way to get a production-scaled (and in the case of a weekly calculation, "cut") curve for the must-run sources. I can look into that.

grdw: > We do still need a way to get a production-scaled (and in the case of a weekly calculation, "cut") curve for the must-run sources. I can look into that.

Cool, than I'll take a look at the influence the heat sources have to the business case calculation.

antw:
Conflits resolved and merged manually. I finished off the heat series for the load chart (the issues/heat_source_load_chart branch) and merged that too.

Issue #1031, opened 2016-06-02
Business cases take very very long to load/calculate


dennisschoenmakers:
Even for simple LES's, it takes at least 40 seconds, at least on my laptop. This is so long, that most users will probably have closed their tabs long before that.

Also, if you refresh the page, you have to wait for another 40 seconds. 😒

antw:
> Also, if you refresh the page, you have to wait for another 40 seconds. 

This is the more immediately fixable issue. It used to be that if the business case had previously been calculated, it would be loaded from the DB without recalculating. I'm not sure when that changed. Perhaps when it moved to a dedicated page?

The initial calculation is slow because it depends on the network calculation running first. #1017 will help in some situations.

dennisschoenmakers:
> I'm not sure when that changed. Perhaps when it moved to a dedicated page?

Anybody know about that?

grdw:
> Anybody know about that?

\*Raises hand\* Yes I do. I'll look into this.

antw:
I think I see what's happening. The request for the `data` json is sending a `clear: true` param every time.

grdw:
Yes, I know 👍 already saw that happening. Am fixing it.

grdw:
@antw I have a question about the 'hackyness' level of a feature. Right now I'm only checking the `cache_updated_at` time for a LES (which is when somebody changes something in the technology matrix) and compare it to the `business_case` updated_at. But the business case should also recalculate when somebody edits his or her strategies, which is now not the case. 

The fast way would be to touch the `cache_updated_at` timestamp when somebody bumps the strategies. This would also make the cache validator simpler. But I'm wondering if that would be 'okido' to do. 

Code wise it would be something like:
````ruby
class SelectedStrategy
...
  after_save :set_testing_ground_cache_updated_at

  ...

  private

  def set_testing_ground_cache_updated_at
     testing_ground.touch(:cache_updated_at)
  end
end
````

antw:
Do you happen to know if that will also invalidate the basic ("strategies off") network cache? I think it will, since `Validator#fresh?` will then return false. That would be a pity.

Could it not instead set something on the BusinessCase instructing it of the need to recalculate? Perhaps it could call `BusinessCase#clear_job!`?

grdw:
> Could it not instead set something on the BusinessCase instructing it of the need to recalculate?

When changing the strategies? I think that's better 👍 I think putting it in the `TestingGround::StrategyUpdater` would be best.

[...]

PR #1034, opened 2016-06-02
Clear business case only when the LES is saved later

grdw:
Part of: #1031


antw:
Worked perfectly; thanks!

Issue #1020, opened 2016-06-02
It hangs when I click Apply


dennisschoenmakers:
To reproduce:

1. Load https://moses.energytransitionmodel.com/testing_grounds/480
5. Click Apply

Indefinitely:
![image](https://cloud.githubusercontent.com/assets/104830/15743412/8cf03734-28c5-11e6-9664-95f9d9ec7934.png)

ChaelKruip:
I think it is definitely a bug that ETMoses 'hangs' but I'm also confused as to why it is possible to click 'apply' at all when you haven't changed anything yet.

Assigning @antw

antw:
It's a display issue only. If you click "Apply" without changing anything, no request is sent to Moses, but for some reason the loading "spinner" is still shown. I'll look into that.

PR #1037, opened 2016-06-02
Fixing toggle of strategies button when strategies are exactly the same

grdw:


ChaelKruip:
Fixes https://github.com/quintel/etmoses/issues/1020

PR #1046, opened 2016-06-03
Destroy network cache before calculating

### LINKED ISSUE #1044: Gas load is not consistent between weeks and sometimes hangs
Unfortunately, I don't have the screenshots to 'prove' it, but this is what I did:

I changed the strategies for https://beta-moses.energytransitionmodel.com/testing_grounds/509

image

I go to the next week

There the gas load seems to be **very** different and inconsistent with the other week.

image

I back to a week earlier

I get console log errors:

image

Then I hit refresh:

It keeps hanging, nothing happens for > 2 minutes

antw:
@grdw Could you see if this fixes the gas cache issue for you?

My guess is that the "strategies off" calculation was finishing before "strategies on", and the app thought that the "strategies on" load was also complete. Thus, it would be sending the *old* cached loads even though the "strategies on" calculation was still running.

This change deletes the cache before starting the calculation. That should prevent the old "strategies on" cache from being considered valid.

Closes #1044

### DIFF
--- a/app/jobs/testing_ground_calculator_job.rb
+++ b/app/jobs/testing_ground_calculator_job.rb
@@ -6,6 +6,7 @@ class TestingGroundCalculatorJob
   end
 
   def perform
+    cache.destroy
     cache.write(network_calculation)
   end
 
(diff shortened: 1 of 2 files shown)


grdw:
Thanks, already saw this flying by in my mail. I'll take a look right now.

grdw:
Works as expected 👍

PR #1061, opened 2016-06-10
Issues/heat asset list

grdw:


ChaelKruip:
@grdw can this be merged?

grdw:
I would say it's ok for now. Otherwise it will be to cluttered with tiny other issues which have no relation to the actual issue.

antw:
@grdw Is there an easy way to convert a serialized value from the DB into something else, before it is displayed in the table? I noticed that `connection_distribution` is stored as a factor (a number between 0 and 1, where 1.0 means 100%), but the front-end labels it as a percentage.

screen shot 2016-06-13 at 14 50 12

I tried creating a custom reader/writer on InstalledHeatAssetLocation...

```ruby
class InstalledHeatAssetLocation
  def connection_distribution_percent
    connection_distribution * 100
  end

  def connection_distribution_percent=(amount)
    self.connection_distribution = amount / 100
  end
end
```

... and then using that in the front-end, but this doesn't work. It looks like the values get saved to the DB without ever hitting the AssetLocation model.

grdw:
>  It looks like the values get saved to the DB without ever hitting the AssetLocation model.

Yes that's correct. My idea for now is to to keep it a `Float`. Eventually when the sliders are implemented (or if you are currently doing that) my idea would be to have a hidden field in the HTML that has the `number_of_heat_connections`. The slider than specifies a number between 0..100 (or 0..1.0) and updates a hidden `number_of_units` input in the table. If that input changes so does the JSON.

antw:
> After thinking some more I guess that at the moment of calculating the business case, the BusinessCaseCalculator class should determine the amount of heat connections. That's passed on to the InitialCosts calculation and than it should be allright I guess?

I agree. I like the idea of storing the percentage in the asset list, rather than the explicit number of units. I'll merge this as-is and we can worry about sliders (or changing the UI to a percentage) once the heat summary chart is done.

Issue #1074, opened 2016-06-13
Legend for individual charts + stacked chart are broken


grdw:

ChaelKruip:
can this be closed @grdw?

PR #1076, opened 2016-06-13
Fix legend for stacked + individual chart

grdw:
Fixes: #1074


grdw:
@antw this currently keeps the Total, Stacked, Invidividual select box disabled. So if you could merge it any time soon that would be great.

PR #1099, opened 2016-06-17
Heat infra yamls

ChaelKruip:
I have added new files for secondary infra.

TODO: the correct values for secondary infra need to be loaded if another type of connection is used.


ChaelKruip:
@grdw can you have a look at the above TODO and then merge this?

grdw:
@ChaelKruip could you test that the heat asset table is working correctly?

ChaelKruip:
> @ChaelKruip could you test that the heat asset table is working correctly?

When I add a new line and specify its type, the numbers do not change (they keep the values of the first line):
etmoses

Issue #914, opened 2016-04-18
[Airbrake] [Production] undefined method `downcase' for nil:NilClass


dennisschoenmakers:
**Airbrake error:** [#9995](https://quintel.airbrake.io/projects/111227/groups/1666507774947209995)
**Airbrake project:** ETMoses

**Error type:** `NoMethodError`
**Error message:** `` undefined method `downcase' for nil:NilClass ``
**Where:** `market_models#show`
**Occurred at:** Apr 18, 2016 08:13:27 UTC
**First seen at:** Apr 18, 2016 08:13:27 UTC
**Occurrences:** 1 (1 since last deploy on Apr 04, 2016 09:57:19 UTC)

**URL:** [https://moses.energytransitionmodel.com/market_models/190](https://moses.energytransitionmodel.com/market_models/190)
**File:** `[PROJECT_ROOT]/app/helpers/market_models_helper.rb`

**Backtrace:**
```
[PROJECT_ROOT]/app/helpers/market_models_helper.rb:29:in format_interaction_foundation
[PROJECT_ROOT]/app/views/market_models/show.html.haml:19:in public_send
[PROJECT_ROOT]/app/views/market_models/show.html.haml:19:in block (2 levels) in _app_views_market_models_show_html_haml___13768061355553179_70327427919640
[PROJECT_ROOT]/app/views/market_models/show.html.haml:13:in each
```

grdw:
Whenever you try to create a market model and submit it empty. It will give this error.

PR #1115, opened 2016-06-20
Correctly validate a market model's interactions

grdw:
Fixes: #914

### DIFF
--- a/spec/models/market_model_spec.rb
+++ b/spec/models/market_model_spec.rb
@@ -3,6 +3,61 @@ require 'rails_helper'
 RSpec.describe MarketModel do
   it { expect(subject).to validate_presence_of(:name) }
 
+  context "when validating" do
+    let(:market) {
+      FactoryGirl.build(:market_model,
+        name: "Test",
+        interactions: [interaction])
+    }
+
+    let(:interaction) {
+      {
+        "stakeholder_from"    => "customer",
+        "stakeholder_to"      => "producer",
+        "tariff"              => 1.0,
+        "tariff_type"         => "fixed",
+        "applied_stakeholder" => "customer",
+        "foundation"          => "abc"
+      }
+    }
+
+    let(:messages) { market.valid? ; market.errors[:interactions] }
+
+    MarketModel::PRESENTABLES.each do |attribute|
+      context "an interaction with no '#{ attribute }'" do
+        let(:interaction) { super().update(Hash[attribute, ""]) }
+        let(:translated_attribute) {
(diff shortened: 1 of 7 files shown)


spec/models/market_model_spec.rb
```diff
@@ -3,6 +3,21 @@
 RSpec.describe MarketModel do
   it { expect(subject).to validate_presence_of(:name) }
 
+  it "validates market model interactions" do
```
antw: This could do with being more specific. Since there are multiple key/values which fail validation, you can't be sure that any one of them are responsible for the failure. I also think that `{"tariff_type" => "fixed"}` is not a valid interaction; the other attributes are required.

Consider something like:

```ruby
RSpec.describe MarketModel do
  context 'when validating' do
    let(:market)      { build(:market_model, interactions: [interaction]) }
    let(:interaction) { valid_interaction }
    let(:messages)    { market.valid? ; market.errors[:interactions] }

    MarketModel::PRESENTABLES.each do |attribute|
      context "an interaction with no '#{ attribute }'" do
        let(:interaction) { super().merge(attribute => '') }

        it 'is not valid' do
          expect(market).to_not be_valid
        end

        it 'has an error on :interactions' do
          expect(messages).to include("value for #{ attribute } can't be blank")
        end
      end

      context "an interaction with a missing '#{ attribute }'" do
        let(:interaction) { super().except(attribute) }

        it 'is not valid' do
          expect(market).to_not be_valid
        end

        it 'has an error on :interactions' do
          expect(messages).to include("value for #{ attribute } can't be blank")
        end
      end
    end # each PRESENTABLES
  end
end
```

... this is way more precise, and more useful for regression testing. For example, if one day a validation on "stakeholder_from" were to stop working, we'll catch it because the test *for that attribute* will fail (and it will be immediately obvious which attribute has a problem). The original test would continue to pass because one of the other attributes would correctly fail the validation.

It also has nice output:

```
$ rspec -fd

MarketModel
  when validating
    an interaction with no 'stakeholder_from'
      has an error on :interactions
      should not be valid
    an interaction with a missing 'stakeholder_from'
      has an error on :interactions
      should not be valid
    an interaction with no 'foundation'
      has an error on :interactions
      should not be valid
    an interaction with a missing 'foundation'
      has an error on :interactions
      should not be valid
    an interaction with no 'tariff'
      has an error on :interactions
      should not be valid
    # ...
```

It would be nice for the message to be [a bit more human-friendly](http://guides.rubyonrails.org/i18n.html#error-message-scopes), or at the very least use `key.humanize.downcase`.

antw:
TY for the changes!

Issue #1170, opened 2016-07-05
Buffering/boosting radio button name incorrect when adding new technologies to newly add buffers


grdw:
How to reproduce:

1. Import a scenario with default topology/market-model
2. Remove all technologies
3. Add 1 buffer to each end-point
4. Add technologies to those buffers
5. Radio buttons begin to turn-off

Issue #1145, opened 2016-06-27
switching concurrency leads to errors


DorinevanderVlies:
In a working LES [(#549)](https://beta-moses.energytransitionmodel.com/testing_grounds/549/) I switched concurrency of buildings and households to minimal.
This resulted in a LES that can not be calculated. Screenshot is shown below.

Switching back to max buildings concurrency does not reset the initial technology profile, but also  results in a LES that can not be calculated.

The same applies when I change the concurrency of households. I have not yet  tried other changing the concurrency of other technologies.

The initial, working, technology profile can be found in the "Ameland" folder in Dropbox. It is named _"TP concurrency bug.csv"_.

![image](https://cloud.githubusercontent.com/assets/19907658/16374613/86d48a76-3c57-11e6-8de7-a9a5e67c3a9b.png)

@dennisschoenmakers

ChaelKruip:
> In a working LES (#549)

For many technologies, the number of units is zero in this LES. I can imagine that the concurrency calculation has trouble dealing with this. Perhaps you could try to remove all of those technologies from the LES and try again? I have confirmed that in a simpler LES the concurrency settings work as expected and result in valid output.

DorinevanderVlies:
Sounds plausible. But I don't think that is the problem.

I have tried again on [LES 553](https://beta-moses.energytransitionmodel.com/testing_grounds/553).
This LES does not contain any technologies with number of units 0.

For any change in concurrency I end up with a LES that cannot be calculated.

**Example**
I set all the PV units to 1. When I change the concurrency of PV to min. Several other technologies are split up and the scenario that is created "could not be calculated".

ChaelKruip:
> I set all the PV units to 1. When I change the concurrency of PV to min. Several other technologies are split up and the scenario that is created "could not be calculated".

I can confirm that his happens. Weird that the PV concurrency slider has impact on the other technologies. @antw can you look into this?

grdw:
I took some time to investigate it a little. 

I tested the following LES's on 'beta':

*With one technology/buffer at one endpoint*:

- A Les with one technology, switching concurrency from min to max with unit > 1. This seems to be working ok. I used one EV/Base load. I don't experience any problems with the calculation. I do see that the demand is empty when adding a new base load but I think this is how it's supposed to work? ( @ChaelKruip )
[...]

PR #1216, opened 2016-07-06
Concurrency bug-squashing

grdw:
Fixes: #1145 
Fixes: #1170


antw:
First impressions are that this is an improvement, but still not quite there yet...

1. I tried to recreate the issue described in https://github.com/quintel/etmoses/issues/1145#issuecomment-230720172:

  > What I found so far is that if I edit the LES and set solar PV concurrency to "min" and hit "Apply" everything seems to work fine. However, if I add a solar PV array to "Endpoint 2" and then change the concurrency setting, the heating technologies on "Endpoint 1" are split up. It seems half the heating technologies are assigned to a non-existent space_heating_buffer_2.

  The good news is that heating technologies no longer get split between the original buffer and a non-existent `buffer_space_heating_2`. However, the buffer does get renamed from `buffer_space_heating_1` to `buffer_space_heating_2`. The buffer number seems to increment by one each time the concurrency settings are toggled and applied.

2. To [the test scenario](https://beta-moses.energytransitionmodel.com/testing_grounds/585), add a single *empty* "Buffer space heating" to "Endpoint 2" and toggle Solar PV concurrency to "min"; heating technologies and buffers are still unchanged at the "max" setting. The heating technologies from the buffer on "Endpoint 1" are *duplicated* onto the buffer on "Endpoint 2" (i.e. if there were 5 units of "space heater combined network gas" on "Endpoint 1", there are now an additional 5 units also installed on "Endpoint 2".

3. Add a second "Buffer space heating" to "Endpoint 1", set demand to `1000` and hit "Apply" on the concurrency settings (no changes made). The original buffer is removed, and all the technologies are moved to the new one.

  I guess that's expected since the concurrency setting for that technology is "max", but for buffers I would expect the demand, volume (and perhaps capacity?) attributes to be summed (i.e. two buffers with demand of 1000 should be merged into one with 2000).

4. Similarly, without making any changes to the test scenario, if "Buffer space heating" is toggled to "min" concurrency, there are now five buffers all with the original amount of demand (total heat demand for the LES is now 5x higher than before).

*3 and 4 also happen on production/beta, so those are not new problems.*

--

This might not be a popular sentiment given the work which has gone into this, but I really think there's a strong case for removing the concurrency settings entirely:

* Using "min" concurrency on anything but the smallest of LESes causes performance problems with the technology form, and results in excessively long network calculations.

* As (3) demonstrates, it's very easy to lose data when switching to "max" concurrency.

  A simpler example: add 2 P2P batteries and set a custom capacity on one. Apply the concurrency settings. You either end up with two custom-capacity batteries, or two default-capacity batteries (depending on which order they appeared on the form).

I would be curious to know how these settings are used in the real-world by our visitors, because I think we could find simpler alternatives for splitting techs into smaller units.

(pinging @ChaelKruip and @dennisschoenmakers)

app/services/testing_ground/concurrency/maximizer.rb
```diff
@@ -0,0 +1,13 @@
+class TestingGround
+  class Concurrency
+    module Maximizer
+      def maximize
+        -> (techs) {
+          tech       = techs.first
+          tech.units = techs.sum(&:units)
+          tech
+        }
```
antw: Very minor thing: use `-> { }` for single-line blocks, and `lambda do ... end` for multi-line blocks.

https://github.com/bbatsov/ruby-style-guide#lambda-multi-line

grdw:
> [..] there are now an additional 5 units also installed on "Endpoint 2".

I just found out point 2 as well .. Working on it as we speak.

grdw:
> but I really think there's a strong case for removing the concurrency settings entirely:

I would have no problem with this personally, but I think that this might be a different problem. 

> LESes causes performance problems with the technology form

Yes, but this is also the case if you have a topology with > 100 end-points. I think we haven't even tried a 1000 end-point topology. The concurrency causes a large form, but it's not the only reason which can cause a large form to render.

> I would be curious to know how these settings are used in the real-world by our visitors, because I think we could find simpler alternatives for splitting techs into smaller units.

Hmm... I get what you are saying here but I wonder what hose simpler alternatives would be? Maybe if the user specifies manually in groups of how many units someone wants to split their technologies?

ChaelKruip:
> This might not be a popular sentiment given the work which has gone into this, but I really think there's a strong case for removing the concurrency settings entirely:

I think we should consider the possibility. I'm not against it in any case.

antw:
> I guess that's expected since the concurrency setting for that technology is "max", but for buffers I would expect the demand, volume (and perhaps capacity?) attributes to be summed (i.e. two buffers with demand of 1000 should be merged into one with 2000).

I don't know what I was thinking when I wrote this. If the two buffers have the same demand, setting `units` to 2 would have to correct effect of doubling the demand. Mind you, that doesn't help if the two buffers have different demands, but that's a another problem entirely.

> Hmm... I get what you are saying here but I wonder what hose simpler alternatives would be? Maybe if the user specifies manually in groups of how many units someone wants to split their technologies?

I think going from min to max concurrency is problematic, since visitors will almost certainly lose data (custom capacities, demands, etc). I would be in favour of a newly-imported scenario being as it is today: max concurrency. Then, if a user desires something more like "min" concurrency, they can hit a "Split" button next to the technology:

screen shot 2016-07-08 at 12 02 14

A technology with `units = 4` becomes two identical technologies with `units = 2`. Then the visitor can customise each one as they desire. If they want, they could then split those into four units of 1.

On the back-end when the LES is saved we would *ideally* merge *identical* technologies together. Two identical solar PV techs could be merged into one with `units = 2`.

grdw:
> they can hit a "Split" button next to the technology:

I kind of like this (how it looks :D ) but what happens after someone has split a technology? How do you return back to normal? Or if you hit split twice in a row, nothing happens?

grdw:
I'd suggest to maybe put this in another issue and keep the current concurrency things as they  are. And maybe later we could actually change it into something better? @ChaelKruip what do you say?

ChaelKruip:
> I'd suggest to maybe put this in another issue and keep the current concurrency things as they are. And maybe later we could actually change it into something better? @ChaelKruip what do you say?

Yes. @grdw can you do this?

antw:
I saw that changes have been pushed, so I tried the [same steps I reported above](https://github.com/quintel/etmoses/pull/1216#issuecomment-231099936):

1. Is fixed!

2. Kind of fixed, but problem (1) makes an appearance. "Buffer space heating 1" on the first endpoint has been renamed to "Buffer space heating 2", and the "Buffer space heating 2" on the second endpoint has been renamed to "Buffer space heating 3"

  Also: Set "Buffer space heating" concurrency to "min" and apply. There are now five buffers (which is correct) with the names "Buffer space heating 6", "Buffer space heating 7", ...

3. The newly-added buffer is renamed (from "2" to "3"), and the original buffer **and attached heating technologies** all disappear.

4. **New**: Set any one of the technologies attached to the buffer to have `units = 1`, "Buffer space heating" concurrency to "min" and apply. There is one instance of the technology attached to the first buffer with `units = 1`. The other four buffers also have an instance of the technology technology with `units = 0`; I don't think those technologies should exist.

If we're not going to turn off the concurrency feature, I still think we should push ahead with merging this. However I do think we should get rid of it ASAP because it's too easy for visitors to shoot themselves in the foot (and a pain for us to maintain and test).

ChaelKruip:
> If we're not going to turn off the concurrency feature, I still think we should push ahead with merging this. 

Let's do this for now.

> However I do think we should get rid of it ASAP because it's too easy for visitors to shoot themselves in the foot (and a pain for us to maintain and test).

I agree but we should present the client with an alternative idea.

antw:
> I agree but we should present the client with an alternative idea.

I'll document my proposal more thoroughly in a new issue later.

I do think (3) needs to be fixed before merging; we can't have entire heat buffers disappearing...

grdw:
It's a little unclear to me what we're going to do with this PR? Can we just close it (without merging) and start with the simple 'split' idea of Anthony any time soon? Or should we try to fix point number 3, merge it and than replace it with the 'split' idea. 

It all comes down a little in which time frame we're going to do the actual replacement. If it's going to take at least 2-3 months before we start with that I'd say, fix (3) and merge it. If it's any time soon than I'd say close it.

CC: @ChaelKruip

antw:
> It all comes down a little in which time frame we're going to do the actual replacement. If it's going to take at least 2-3 months before we start with that I'd say, fix (3) and merge it. If it's any time soon than I'd say close it.

The sprint plan includes "Concurrency-code should be replaced by something less error-prone", so hopefully we might get to it this month? I'll post an issue describing the "Split" button I proposed and we can discuss it.

If it is indeed going to be more than a month, I agree that we should merge this after fixing (3), because it still works much better than the old concurrency code.

ChaelKruip:
> The sprint plan includes "Concurrency-code should be replaced by something less error-prone", so hopefully we might get to it this month? I'll post an issue describing the "Split" button I proposed and we can discuss it.

👍 

I think we should close this and start from scratch soon!

PR #1232, opened 2016-07-08
Issues/#1134 heat infra tlc

grdw:


grdw:
I'm closing this, for this needs to be replaced by sliders and not be solved by something this patchy...

PR #1237, opened 2016-07-08
Updating business case calculation for initial costs

grdw:
Make sure that `depreciation_costs` are calculated in the same fashion accross different technologies from different infrastructuers.

### DIFF
--- a/app/models/installed_gas_asset.rb
+++ b/app/models/installed_gas_asset.rb
@@ -1,29 +1,34 @@
 class InstalledGasAsset
   include Virtus.model
+  include BusinessCaseCosts
 
   attribute :pressure_level_index, Integer
   attribute :part, String
   attribute :type, String
-  attribute :amount, Float, default: 1
+  attribute :units, Float, default: 1
   attribute :stakeholder, String, default: 'system operator'
   attribute :building_year, Integer
-  attribute :lifetime, Integer
-  attribute :investment_cost, Float
+  attribute :technical_lifetime, Integer
+  attribute :initial_investment, Float
 
   def decommissioning_year
-    building_year + lifetime
+    building_year + technical_lifetime
   end
 
   def net_present_value_at(year)
     if year.between?(building_year, decommissioning_year)
-      ((lifetime - (year.to_f - building_year.to_f)) / lifetime) * investment_cost
+      ((technical_lifetime - (year.to_f - building_year.to_f)) / technical_lifetime) * initial_investment
     else
       0
     end
   end
 
   def total_investment_costs
-    investment_cost * amount
+    initial_investment * units
+  end
+
+  def om_costs_per_year
+    part_record.yearly_o_and_m_cost
   end
 
   # Public: The pressure level (in bars) to which the asset is assigned.
--- a/db/migrate/20160714083258_update_gas_asset_lists.rb
+++ b/db/migrate/20160714083258_update_gas_asset_lists.rb
@@ -1,0 +1,22 @@
+class UpdateGasAssetLists < ActiveRecord::Migration
+  def change
+    translations = {
+      investment_cost: :initial_investment,
+      lifetime: :technical_lifetime,
+      amount: :units
+    }
+
+    GasAssetList.all.each do |gas_asset|
+      asset_list = gas_asset.asset_list.map do |asset_item|
+        translations.each_pair do |old_key, new_key|
+          asset_item[new_key] = asset_item[old_key]
+          asset_item.delete(old_key)
+        end
+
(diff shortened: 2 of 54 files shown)


ChaelKruip:
@grdw is this ready to be looked at by @antw?

app/models/installed_gas_asset.rb
```diff
@@ -10,6 +11,10 @@ class InstalledGasAsset
   attribute :lifetime, Integer
   attribute :investment_cost, Float
 
+  alias_method :initial_investment, :investment_cost
+  alias_method :technical_lifetime, :lifetime
+  alias_method :units, :amount
```
antw: Maybe it would be worthwhile to rename these attributes to match those in Technology?

app/models/network/node.rb
```diff
@@ -1,5 +1,7 @@
 module Network
   class Node < Turbine::Node
+    include BusinessCaseCosts
```
antw: This stuff doesn't belong here. :smile:

The `Network` namespace is for calculating load. Anything relating to costs should be part of the `Market` module.

grdw:
I'll fix these points 👍 thanks for notifying.

app/models/network/node.rb
antw: If what you want to do it take a list of nodes and ask them for their costs, while that is not the responsibility of a *Network::Node*; the [Decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) would allow you to do this while keeping your other code simple:

```ruby
class NodeCostsDecorator
  include BusinessCaseCosts

  def initialize(node)
    @node
  end

  def initial_investment
    # ...
  end

  def om_costs_per_year
    # ...
  end

  # ...
end
```

Then, instead of using the Network::Node instances directly, you wrap each node in an Decorator and use those instead:

```ruby
nodes = network.nodes.map { |node| NodeCostsDecorator.new(node) }
```

You can see this in action in TestingGround::HeatSummary [where TechnologySummary](https://github.com/quintel/etmoses/blob/7a9c55e4b1b0da1c5fb8e0bd084e67e0824c6631/app/services/testing_ground/heat_summary.rb#L172-L197) takes `Network` objects, and then create a new interface for use in the HeatSummary without polluting the original class with new methods.

grdw: I was already making a Decorator haha 😅 . Thanks for explaining though.

grdw:
@antw I fixed your suggested changes. If you could have another look.

antw:
waynes-world_excellent

db/migrate/20160714083258_update_gas_asset_lists.rb
```diff
@@ -0,0 +1,22 @@
...
+      investment_cost: :initial_investment,
+      lifetime: :technical_lifetime,
+      amount: :units
+    }
+
+    GasAssetList.all.each do |gas_asset|
+      asset_list = gas_asset.asset_list.map do |asset_item|
+        translations.each_pair do |old_key, new_key|
+          asset_item[new_key] = asset_item[old_key]
+          asset_item.delete(old_key)
```
antw: FYI: `Hash#delete` returns the deleted value, or `nil` if the given key doesn't exist. So you can do it all in one operation:

```ruby
{ a: 1 }.delete(:a) # => 1
{ a: 1 }.delete(:b) # => nil

new_thing = { a: 1 }.delete(:a) # => 1
```

antw:
I have reverted this on account of #1268, #1269, and #1270, and the branch `issues/business_case_initial_costs` has been restored.

grdw:
> I have reverted this on account of #1268, #1269, and #1270, and the branch issues/business_case_initial_costs has been restored.

Aren't these errors just triggered because of someone being busy at the same time as the deploy? Or?

antw:
> Aren't these errors just triggered because of someone being busy at the same time as the deploy?

I don't think so. I wouldn't expect a method to disappear in the middle of a DelayedJob run.

*  The [gas asset table still refers to `amount`](https://github.com/quintel/etmoses/blob/issues/business_case_initial_costs/app/views/gas_asset_lists/_table.html.haml#L28) (#1268)
* InstalledHeatAssetPipe is missing `total_yearly_costs`, [called here](https://github.com/quintel/etmoses/pull/1237/files#diff-daddc539cfb1e6eb86493978271385a4R17) (#1269):

 ```ruby
   InstalledHeatAssetPipe.new.respond_to?(:total_yearly_costs)
  ```

* #1270 is just the standard "business case failed to calculate" error, so no worries there.

grdw:
#1270 is a result of #1269

Issue #1178, opened 2016-07-05
"Save all and view gas/heat load" button does not save changes


ThomBuijs:
Why can't I alter the distance and associated with source settings for the primary network heat asset list

antw:
Looks like the gas and heat forms are not marked as "editing".

@ThomBuijs The blue "Save" buttons will work in the meantime.

PR #1241, opened 2016-07-11
Marks editable table forms as edtiable when changes arise

grdw:
Fixes: https://github.com/quintel/etmoses/issues/1178


antw:
Pull request looks good, and *all* fields now appear to be saved correctly. Great!

One small issue however: the heat-source must-run profile box remains disabled:

| `master` | `issues/#1178-save-all` |
| --- | --- |
| screen shot 2016-07-11 at 12 50 11 | screen shot 2016-07-11 at 12 51 36 |

grdw:
Yes, I know where this comes from :+1: I'll look into it.

grdw:
I fixed the profile problem. The thing is that when you changed a value in the heat assets matrix the changes got overwritten because of the [callback](https://github.com/quintel/etmoses/blob/master/app/assets/javascripts/tables/heat_asset_list_table_part.js#L8-L20). This is now solved by not initiating the callback upon page load but only when you change the actual technology.

PR #1272, opened 2016-07-14
Updating business case calculation for initial costs

grdw:
Make sure that `depreciation_costs` are calculated in the same fashion accross different technologies from different infrastructuers. 

Attempt 2


grdw:
@antw I fixed the two renaming problems and tested it manually. This should work fine now 😅

PR #1276, opened 2016-07-15
Clean up gas chart + initially hide the yearly gas chart

grdw:


app/assets/javascripts/graph.js
```diff
@@ -0,0 +1,53 @@
+window.graphs = {};
+
+var Graph = (function () {
```
antw: I'm a bit late to the party in bringing this up, but... these are *charts* rather than graphs. I know that sounds like nitpicking, but we regularly use the term "graph" to refer to the data structure, and "chart" to mean front-end visuals, and ETModel is consistent in the use of "chart". I think it risks confusion to mix these two terms up.

antw: I see the misuse of "graph" is quite widespread. 😆 

Never mind; I'll merge this and open a low-priority issue to adjust the names later.

PR #1293, opened 2016-07-18
Add a delete button for a LES

grdw:
- Adds a destroy method to a LES
- Cleaning up destroy methods for Topology, Market Model

### DIFF
--- a/app/controllers/market_models_controller.rb
+++ b/app/controllers/market_models_controller.rb
@@ -49,11 +49,7 @@ class MarketModelsController < ResourceController
   end
 
   def destroy
-    if TestingGround.where(market_model: @market_model).count > 0
-      @market_model.update_attribute(:user, User.orphan)
-    else
-      @market_model.destroy
-    end
+    TestingGround::Destroyer.destroy_part(@market_model)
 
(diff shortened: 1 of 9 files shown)


app/policies/market_model_policy.rb
```diff
@@ -1,6 +1,10 @@
 class MarketModelPolicy < ApplicationPolicy
   include PrivatePolicy
 
+  def destroy?
+    super && TestingGround.where(market_model: record).count <= 1
```
antw: Instead of `<= 1` – which would theoretically allow a market model to be deleted while one LES was still using it – could the LES be deleted *first*, and then just check if the count is zero?

I'm actually a bit on the fence about mixing the Policy/authorization stuff into the TestingGround model, I feel like that's something which belongs at the level of the controller (or a service object).

Perhaps this might be encapsulated as a `TestingGround::Destroy` service (like [`SaveAs`](https://github.com/quintel/etmoses/blob/master/app/services/testing_ground/save_as.rb))?

antw:
Merged manually to resolve conflicts. :+1:

Issue #1292, opened 2016-07-18
It should not be possible to set the efficiency/COP for technologies that don't support it


ChaelKruip:
We should 'disable' it in the interface but still show it for now.

antw:
Already on it. :grinning:

ChaelKruip:
Can this be closed @antw?

PR #1298, opened 2016-07-18
Show only whitelisted attributes on technology form

antw:
This is note *quite* what was asked for in #1292, because it *hides* COP – rather than disables it – for technologies on which it is not supported. I think that's an acceptable compromise in the name of simplicity, since you can still tell that efficiency is 100% by comparing the input and output capacities. If you feel strongly that it should be visible, let me know. I intend to revisit this with "Input Validation".

I have also hidden `demand`, `volume`, and capacity attributes in some cases. For example, technologies which belong to heat buffers should not allow the visitor to set a `demand` or `volume` (those are set on the buffer). Other technologies, such as Households, have their load set by ``demand`` and I felt that also showing a capacity was confusing (and the two attributes would conflict).

* Removed `demand` and `volume` from all heating technologies.

* Removed COP from all technologies except:
  * Heat pumps
  * Hybrid heat pumps
  * Power-to-heat

* From base loads, removed capacity and `volume`.

* From P2G, removed `demand`.

* From the neighbourhood battery, removed `demand`.

* From power-to-heat, removed `demand`.

* P2P batteries, removed `demand`.

* From solar PV, removed `demand` and `volume`.

* From the electric vehicle, removed `demand`.

@ChaelKruip I'd appreciate a critical eye over these changes; I don't think they'll be controversial, but just in case...

### DIFF
--- a/app/assets/stylesheets/testing_grounds.css.scss
+++ b/app/assets/stylesheets/testing_grounds.css.scss
@@ -442,6 +442,10 @@ table.table.scenarios{
         float: left;
       }
 
+      &.carrier_capacity {
+        clear: left;
+      }
+
       &.buffer select,
       &.profile select{
         width: 550px;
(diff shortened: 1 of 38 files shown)


antw:
screen shot 2016-07-18 at 16 53 55

Issue #1179, opened 2016-07-05
It should be possible to clone a public LES and edit it yourself


MarjoleinSchwachofer:
Instead of creating a whole new LES by going through ETM, we would like to be able to clone an existing LES and make some minor changes and see the results and compare this. Currently you can only see a public case and see the electricity load, gas load and businesscase. Bu you cannot see the gas assets and the like and change one item or just add one node.

ChaelKruip:
> Instead of creating a whole new LES by going through ETM, we would like to be able to clone an existing LES and make some minor changes and see the results and compare this. 

This is problematic if any part of the original LES contains private information. Currently several parts of the LES (say, the Market Model, the Topology, the Gas-Assets etc.) can be either public or private. Only if a LES is completely public, would it be possible to 'clone' it and save it as your own new LES.

In [this issue](https://github.com/quintel/etmoses/issues/839) we discuss the idea of 'merging' the various parts of a LES (Market Model, the Topology, the Gas-Assets etc.) to alleviate the issue of having to keep track of all these components. 

Moreover, we suggest to make three or more 'default' lesses which the user can open, explore, clone and re-use easily to help them make a quick start.

MarjoleinSchwachofer:
ik voeg jullie toe aan de discussie ter info: @Sanne512 @ThomBuijs @JeroenvdLogt

ChaelKruip:
> ik voeg jullie toe aan de discussie ter info: @Sanne512 @ThomBuijs @JeroenvdLogt

Friendly reminder: please use English as we are working with an international crowd here :)

grdw:
About duplicating a LES:

A les can be duplicated if the LES is public, the accompanied Market Model is public and the Topology is public. The duplication will duplicate all parts (gas asset list, heat asset list, heat source list, business case and the selected strategy). The duplication will point to the same Topology and the Market Model as the original LES.

````
LES       DUPLICATED LES
|          /
|         /
TOPOLOGY
MARKET_MODEL
````

When somebody makes a change to either the Topology or the Market model the following will happen.

````
LES       DUPLICATED LES
| |       |        \
| |       |          \
|TOPOLOGY |   TOPOLOGY CHANGED
|         |
MARKET_MODEL
````

LES and DUPLICATED LES are interchangeable in the above example, as in, if someone makes a change to a Topology or Market Model that's already in use with another LES (which is the case when it's duplicated for both the duplicated LES adn the original) the Topology and the Market Model will clone themselves when edited.

PR #1305, opened 2016-07-20
Adding a basic LES duplicator

grdw:
Fixes: #1179

### DIFF
--- a/app/controllers/testing_grounds_controller.rb
+++ b/app/controllers/testing_grounds_controller.rb
@@ -1,6 +1,6 @@
 class TestingGroundsController < ResourceController
   RESOURCE_ACTIONS = %i(edit update show technology_profile data destroy save_as
-                        update_strategies gas_load heat_load)
+                        update_strategies gas_load heat_load clone)
 
   respond_to :html, :json
   respond_to :csv, only: :technology_profile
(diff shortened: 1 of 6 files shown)


app/services/testing_ground/duplicator.rb
```diff
@@ -0,0 +1,35 @@
+class TestingGround::Duplicator
```
antw: This class feels a *lot* like [`TestingGround::SaveAs`](https://github.com/quintel/etmoses/blob/master/app/services/testing_ground/save_as.rb) to me. Did you see that class? 😀

grdw: > Did you see that class?

Well no I didn't 👍 thanks for pointing that out.

ThomBuijs:
I was looking for the status of #1179 and was referred to this issue. Is it closed and fixed?

antw:
> Is it closed and fixed?

It is.

screen shot 2016-08-11 at 18 08 47 means that the change has been accepted, and is published to the "beta" server automatically within a couple of minutes.

antw:
screen shot 2016-08-11 at 18 11 25

You may only clone a LES which is "public" (anyone can view it). If you have a LES which isn't public, you can create a clone by choosing "Edit LES", clicking the "General" tab near the top of the page, and then "Save as new Local Energy Solution".

Issue #993, opened 2016-05-24
[Airbrake] [Production] key not found: "buffer_space_heating_16"


dennisschoenmakers:
**Airbrake error:** [#6619](https://quintel.airbrake.io/projects/111227/groups/1692728902015236619)
**Airbrake project:** ETMoses

**Error type:** `KeyError`
**Error message:** `` key not found: "buffer_space_heating_16" ``
**Where:** `<no information>`
**Occurred at:** May 24, 2016 12:29:18 UTC
**First seen at:** May 24, 2016 12:29:18 UTC
**Occurrences:** 1 (1 since last deploy on May 24, 2016 09:55:58 UTC)

**URL:** `<no information>`
**File:** `[PROJECT_ROOT]/app/models/calculation/technology_load.rb`

**Backtrace:**
```
[PROJECT_ROOT]/app/models/calculation/technology_load.rb:88:in fetch
[PROJECT_ROOT]/app/models/calculation/technology_load.rb:88:in composite
[PROJECT_ROOT]/app/models/calculation/technology_load.rb:48:in block (2 levels) in techs_for
[PROJECT_ROOT]/app/models/installed_technology/profile_curve.rb:17:in block in each
```

dennisschoenmakers:
*Comment from Airbrake:* Resolved [error](https://quintel.airbrake.io/projects/111227/groups/1692728902015236619) reoccurred.

Issue #1321, opened 2016-07-25
Heating technologies should get separate buffers with correct number of units by default


ChaelKruip:
Probably after [this PR](https://github.com/quintel/etmoses/pull/1277), the importing of heating technologies adds all technologies to the same buffer and gives them the 'units' of that buffer:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/17106968/98de0868-528d-11e6-9f4a-792987302f4c.png)
 
[This comment](https://github.com/quintel/etmoses/issues/1285#issuecomment-234235507) has not been incorporated in the implementation leading to many more technologies being created than present in the original ETM scenario.

ChaelKruip:
Closing this as the the related PR has been reverted and this issue no longer occurs.

Issue #1285, opened 2016-07-18
What should the 'split'-function do for buffers?


ChaelKruip:
In general I think that splitting a buffer should to the following:

* Split the buffer into X buffers (as long as the 'units' of the parent buffer > X and the number of available distinct profiles > X)
* Connect the technologies which were connected to the parent buffer to the child buffers as fairly as possible

Specifying the second bullet is the aim of this ticket.

# Case one: same technologies, units = buffer_units

*Starting situation*:
* buffer_1: units = 2
* technology_1: units = 2, connected to buffer_1

*Action*: split into maximum number of groups

*Resulting situation*:
* buffer_1: units = 1
* technology_1: units = 1, connected to buffer_1
* buffer_2: units = 1
* technology_2: units = 1, connected to buffer_2

# Case two: same technologies, units > buffer_units

*Starting situation*:
* buffer_1: units = 2
* technology_1: units = 5, connected to buffer_1

*Action*: split into maximum number of groups

*Resulting situation*:
* buffer_1: units = 1
* technology_1: units = 3, connected to buffer_1
* buffer_2: units = 1
* technology_2: units = 2, connected to buffer_2

# Case three: mixed technologies, units = buffer_units

*Starting situation*:
* buffer_1: units = 2
* technology_1: units = 2, connected to buffer_1
* technology_2: units = 2, connected to buffer_1

*Action*: split into maximum number of groups

*Resulting situation*:
* buffer_1: units = 1
* technology_1: units = 1, connected to buffer_1
* technology_2: units = 1, connected to buffer_1
* buffer_2: units = 1
* technology_1: units = 1, connected to buffer_1
* technology_2: units = 1, connected to buffer_1

grdw:
Copy pasting my content from issue #1283 

The splitting up of buffers is harder than first imagined. So we're going to give it a shot in this issue. How I'm currently thinking about this problem is as follows:

#### Scenario 1 (1 Buffer, 1 technology attached to buffer, units similar)

````
Buffer #1 (2)
↳ HP Water (2)
````
After split in groups of 2:

````
Buffer #1 (1)
↳ HP Water (1)
Buffer #2 (1)
↳ HP Water (1)
````

----

#### Scenario 2 (1 Buffer, 1 technology attached to buffer, units technology lower than buffer)

````
Buffer #1 (2)
↳ HP Water (1)
````
After split in groups of 2:

````
Buffer #1 (1)
↳ HP Water (1)
Buffer #2 (1)
````

[...]

Issue #1267, opened 2016-07-14
Discussion: Add button to "split" technologies; replace concurrency


antw:
[As suggested here](https://github.com/quintel/etmoses/pull/1216#issuecomment-231333920), I'd like to propose a "Split" button as a replacement for the concurrency options we have today.

The idea is that any technology with two or more `units` can be split into two separate technologies, with half the number of units each.

For example, if we have a Household with 4 units:

> <img width="895" alt="screen shot 2016-07-14 at 10 35 51" src="https://cloud.githubusercontent.com/assets/4383/16835076/cb22938a-49ae-11e6-8f9c-f14688e3c753.png">

... the visitor may hit "Split", which will result in two households each containing two units:

> <img width="895" alt="screen shot 2016-07-14 at 10 36 58" src="https://cloud.githubusercontent.com/assets/4383/16835094/ebc84fee-49ae-11e6-9ff7-6bd1603f8b42.png">

If that is not enough, the visitor can then click "Split" on *those* two, ending up with four separate technologies, each with one unit:

> <img width="896" alt="screen shot 2016-07-14 at 10 39 16" src="https://cloud.githubusercontent.com/assets/4383/16835174/58469c98-49af-11e6-8905-fe0d8df40bdc.png">

The "Split" button would be greyed out and disabled if the technology has only one unit installed.

--

Things we might consider:

* Should the split technologies both use the same profile, or should we auto-select a different profile for the new technology?

* Should we support splitting buffers and all their technologies? (I think for simplicity, no).

* *Nice-to-have?* Should we allow a user customise how many technologies are created? Perhaps they want to split a `units=50` technology into five separate technologies of 10 units? That could be a "..." option which the user clicks, and enters how many techs they'd like (up to, but not exceeding, the number of units already installed):

  <img width="265" alt="screen shot 2016-07-14 at 11 15 20" src="https://cloud.githubusercontent.com/assets/4383/16836262/9c961e00-49b4-11e6-843c-35ed5de95422.png">

  I'm on the fence about this, I would like to avoid adding a "..." button – at least at first – and see if a simple "Split" button will suffice. We could add "..." if necessary later.

#### Motivation

1. Splitting technologies using the current "min concurrency" option is like using a sledgehammer. Maybe the user only wants to split that solar array of 500 units into two or three separate technologies, not 500?

2. Discouraging huge numbers of `units=1` technologies will result in faster calculations, and a more responsive technology form.

[...]

PR #1322, opened 2016-07-25
Issues/concurrency split

antw:
Reopened version of #1277 due to #1321.

> What this pull request is and what is isn't:
> 
> *What it is:*
> A basic setup of the technology splitter as described in: #1267. It splits a technology in groups of 2. The code however is build in a way that it can split into groups [larger than 2](https://github.com/quintel/etmoses/pull/1277/files#diff-b32f238bbcae4cdd5ef98a07b151e748R73) however there's currently not yet an interface available to do this. This pull request also contains lots of removal of the old concurrency code. The code also is able to split buffers and their children. The children of buffers have no units as described [here](https://github.com/quintel/etmoses/issues/1285#issuecomment-233363883).
> 
> *What it isn't:*
> The splitter doesn't automatically selects new profiles for the cloned templates. @ChaelKruip and I decided not to do this since the behavior (when cloning) is not how we expected it to be.

----

Fixes: #993, #998, #999, #1120, #1145, #1203, #1368, #1379, #1382


grdw:
Waiting to fix #1368 before merging this. Will commit the result in this same PR.

antw:
I gave this a spin, and the "split" feature and HHP both seem to work nicely.

However I'm concerned about how we will deal with existing LESes. For example, if you have a buffer in a LES like this:

* Space heating buffer – Units: 100
  * Heat pump (air) – Units: 30
  * Electric heater – Units: 70

... and then re-save the LES, you end up with:

* Space heating buffer – Units: 100
  * Heat pump (air) – Units: 100
  * Electric heater – Units: 100

A newly-imported LES would of course have two separate buffers:

* Space heating buffer 1 – Units: 30
  * Heat pump (air)
* Space heating buffer 2 – Units: 70
  * Electric heater

--

I think we might need to add a migration to convert old LESes to the new way of doing things, otherwise costs and loads are going to be dramatically different.

grdw:
> I think we might need to add a migration to convert old LESes to the new way of doing things, otherwise costs and loads are going to be dramatically different.

Yes. So in fact; grab every child/technology from every buffer. Create a new buffer with the units of the child/technology and attach it to the child/technology.

grdw:
I've managed to write a migration for all the old LES's [here](https://github.com/quintel/etmoses/pull/1322/files#diff-221897ffc2140123e1acf5e9b383d3d1R1) 👍 

@antw If you could try it out.

antw:
I pushed a change to the migration to ensure each buffer got a unique index (otherwise it was starting at "1" for each endpoint; it could result in several "buffer_space_heating_1" in a LES). Otherwise, the code looks fine and the new features work as expected. :+1:

I'm holding off on merging (briefly) while I check out differences in heat demand and supply between `master` and this branch:

| Branch | Chart |
| :-: | --- |
| master | screen shot 2016-08-23 at 12 18 11 |
| PR | screen shot 2016-08-23 at 12 18 05 |

This is a default NL/2050 scenario. I'm a bit bothered by the deficit (which seems proportional to demand) and the *almost* total absence of buffering. Once I've managed to explain these two things, and I'm sure they're not caused by a bug in the calculation, I'll merge the pull request.

antw:
### Deficit

The deficit is caused by a buffer whose technology is a district heat connection. The default scenario has no dispatchable central heat production, and must-runs were being imported without being assigned a number of units or a profile. 😕 

[I have resolved this](https://github.com/quintel/etmoses/commit/9233b00290ba5612eb5efa9f4f016d1596e933db); any deficit which still exists in the default scenario is due to insufficient supply from the central heat network.

### Lack of buffering

In the original scenario there is no excess PV, therefore heat is buffered by heat pumps only. A single heat pump was connected to a buffer with a myriad of other technologies. That buffer had a volume of 2.9065kWh and 34 units, for a total volume of slightly under 100kWh. The heat pump therefore runs for a long time to fill up the buffer.

In the new scenario, the heat pump is connected to a buffer with only 1 unit. Therefore it fills the buffer up much sooner, and then turns off. Therefore, I think the lack of buffering is correct.

grdw:
Does this need to any action from me before this can be merged?

antw:
Thanks @grdw! I merged this since I think the buffering behaviour is correct.

PR #1355, opened 2016-08-08
Info in front end

ChaelKruip:
Adding some lines of info and links to the wiki to the tabs of the LES.

### DIFF
--- a/app/views/business_cases/_form.html.haml
+++ b/app/views/business_cases/_form.html.haml
@@ -1,3 +1,13 @@
+%h2 Business case matrix
+%p The matrix below summarizes three types of annual cash-flows
+%ol
+  %li <b>Depreciation costs</b>: the investment of technologies and infrastructure divided by their technical lifetime
+  %li <b>Fixed operation and maintenance costs</b>: defined for technoligies and infrastructure
+  %li <b>Market model rules</b>: transactions between stakeholders as defined in the Market model tab
+
+Item 1 and 2 are summed and displayed on the diagonal of the matrix. The transactions between different stakeholders are shown in the other cells. Rows add up to total (annual) revenu and columns add up to total (annual) expenses. A correction for unmodelled costs (for example removal costs of infrastructure) can be included with the 'free form' row.
+More information can be found in the <a href='https://github.com/quintel/etmoses/wiki/Business-cases'>documentation</a>.
+
 .table-wrapper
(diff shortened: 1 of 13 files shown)


app/views/business_cases/_form.html.haml
```diff
@@ -1,3 +1,14 @@
+%h2 Business case matrix
+.help-block
+  %p The matrix below summarizes three types of annual cash-flows
+  %ol
+    %li <b>Depreciation costs</b>: the investment of technologies and infrastructure divided by their technical lifetime
```
antw: Minor gripe, but [go with `strong` instead of `b`](http://stackoverflow.com/a/271776). You can eliminate the HTML by doing:

```haml
%li
  %strong Depreciation costs:
  the investment of technologies and infrastructure divided by their technical lifetime
```

app/views/heat_asset_lists/_primary_table.html.haml
```diff
@@ -1,4 +1,6 @@
-%h2 Primary network
+%h2 Primary heat infrastructure
+.help-block
+  %p The table below lists the transport pipes between your sources and the feed-in point of the heat network in your LES. You can edit the cost aspects of this primary infrastructure here. The lengths of pipes are determined by the distance of the sources in the Heat source tab. More information can be found in the <a href='https://github.com/quintel/etmoses/wiki/Heat-network'>documentation</a>.
```
antw: If possible, aim for a maximum of 80 characters per line. You can split the content across multiple lines by intending another level:

```haml
.help-block
  %p
    The table below lists the transport pipes between your sources and the
    feed-in point of the heat network in your LES. You can edit the cost
    aspects of this primary infrastructure here. The lengths of pipes are
    determined by the distance of the sources in the Heat source tab. More
    information can be found in the documentation.

```

app/views/gas_asset_lists/_table.html.haml
```diff
@@ -1,3 +1,15 @@
+%h2 Gas infrastructure
+.help-block
+  %p
+  In the table below you can edit the specifics of the gas infrastructure of your LES. Three types of network components can be chosen:
```
antw: Needs another two spaces in front of the text `In the table...`, so that the content belongs to the `%p` above.

```haml
.help-block
  %p
    In the table below you can edit the specifics of the gas in [...]
```

Issue #1318, opened 2016-07-25
Stacked load should be ordered according to variability/spikiness


ChaelKruip:
Stacked plots are difficult to read because curves with high variability are stacked under more well-behaved ones:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/17104051/64448872-5281-11e6-8e50-fe18166937ab.png)

A possible technique to improve this is to order the curves according to how spiky they are. A measure for spikiness could be the following:
```
spikiness = MAX(curve) / SUM(curve)
```
The curves should be ordered beginning with the lowest spikiness up to the highest spikiness.

grdw:
I copied the 'spikiness' method from ETModel and implemented it in ETMoses.

PR #1370, opened 2016-08-12
Sort series by spikiness instead of the label name

grdw:
Fixes: #1318

### DIFF
--- a/app/assets/javascripts/graphs/d3_load_chart.js
+++ b/app/assets/javascripts/graphs/d3_load_chart.js
@@ -123,10 +123,10 @@ var D3LoadChart = (function () {
 
     function setArea(d, scope) {
         if (d.visible && d.area) {
-            if (!(this.settings.view_as === 'stacked')) {
-                return chartParts[scope].colorArea(d.values);
-            } else {
+            if (this.settings.view_as === 'stacked') {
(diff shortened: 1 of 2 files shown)


antw:
I wonder if there's something going on with my dev env, because like #1373 I'm not seeing any difference from master. Although the sort function is definitely being called, I can have it return any random values and the series always appear in the same order.

I tested using a local copy of [beta-LES 668](https://beta-moses.energytransitionmodel.com/testing_grounds/668), "Households 1":

```js
return Math.abs((d3.max(values) - d3.min(values)) / d3.sum(values));
```

screen shot 2016-08-17 at 09 43 06

```js
// reverse order
return -Math.abs((d3.max(values) - d3.min(values)) / d3.sum(values));
```

screen shot 2016-08-17 at 09 43 28

Could you try that LES and see if you get a different result to me?

grdw:
With this one I was already doubting if my sorting actually did something or not... It's weird since sort actually is being called. And `sort()` is destructive so I don't need to re-assign it to the `data` variable again. Hmm.. I need some more time to figure out what's going on here.

grdw:
Turns out that the `sort()` method was initiated to late ... I'll spare the details. 😅 

It should be working now!

Issue #1356, opened 2016-08-09
Gas load chart should show selected week when date picker is set by electricity load chart


ChaelKruip:
When you select a week in the electricity load chart, that week is also selected for the gas load chart. Neat!

The problem is that the chart itself does not display that week (but rather the first week of the year):
![etmoses](https://cloud.githubusercontent.com/assets/1303760/17510599/f0145414-5e1f-11e6-81c3-ec697520b07d.png)

PR #1373, opened 2016-08-15
Set date select week for gas chart correctly

grdw:
Fixes: #1356

### DIFF
--- a/app/assets/javascripts/graphs/d3_base_static_load_chart.js
+++ b/app/assets/javascripts/graphs/d3_base_static_load_chart.js
@@ -3,13 +3,22 @@
 var D3BaseStaticLoadChart = (function () {
     "use strict";
 
-    var spinner;
+    function drawChart(params, firstTime) {
+        Ajax.json(this.settings.url, params, function (data) {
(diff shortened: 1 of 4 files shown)


antw:
Maybe I'm doing something wrong, but when I load the gas chart, having selected the second week in the electricity chart, it's still loading data for the *first* week of the year (even though the date select says otherwise):

```json
{"calculation":{"range_start":0,"range_end":672,"resolution":"high"}}
```

antw:
Whenever I select a new chart (Electricity → Gas, Heat → Electricity), it resets back to week 1. Is that by design? It's still better than labelling the data with the wrong week, so although it's a pity to lose the ability to remember which week was selected, we can merge this if it's intentional and @ChaelKruip is happy.

ChaelKruip:
> Whenever I select a new chart (Electricity → Gas, Heat → Electricity), it resets back to week 1.

I would be happier if it would remember the week but if that is very difficult I can live with resetting.

grdw:
>  I can live with resetting.

Fine. The normal behavior in the first place would be resetting. I'll make a separate issue for the synching.

PR #1384, opened 2016-08-22
Fixes: #540

grdw:
Raise an error when the financials are blank. As explained in #540 the problem arises from another calculation error. The error is usually visible in the DelayedJob queue.

### DIFF
--- a/app/services/finance/business_case_summary.rb
+++ b/app/services/finance/business_case_summary.rb
@@ -7,6 +7,8 @@ module Finance
     end
 
     def summarize
+      return [] if financials.nil?
+
       financials_without_freeform.each_with_index.map do |financial_rule, index|
         summarize_rule(financial_rule, index)
       end
--- a/app/views/business_cases/_table.html.haml
+++ b/app/views/business_cases/_table.html.haml
@@ -1,14 +1,11 @@
 %div.compare
-  - if business_case_rows
+  - if business_case_rows && business_case_rows.size > 0
     %span= "Compare business case from #{@testing_ground.name} with"
     %span.compare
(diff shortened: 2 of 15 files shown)


app/services/finance/business_case_summary.rb
```diff
@@ -7,6 +8,8 @@ def initialize(business_case)
     end
 
     def summarize
+      raise BusinessCaseCalculationError if financials.nil?
```
antw: I don't really understand the purpose behind the exception here. It feels like we're using an exception for flow control (i.e., as a GOTO).

We could eliminate the exception (and the `rescue` in the controller) by doing:

```ruby
return [] if financials.nil?
```

... and nothing in the view has to change.

app/views/business_cases/_table.html.haml
```diff
@@ -58,6 +55,15 @@
     %span.loading-spinner
       = image_tag 'wait.gif'
 
+  - elsif business_case_rows.empty?
+    .panel.panel-danger
+      .panel-heading Something went wrong with the business case calculation.
+      .panel-body
+        Please submit this error to the ETMoses&nbsp;
+        %a{ href: 'https://github.com/quintel/etmoses/issues' }> issues page
```
antw: Perhaps instead of relying on the visitor to report the error we might find a way to forward DelayedJob exceptions to Airbrake?

grdw: > we might find a way to forward DelayedJob exceptions

Yes. Maybe [this](https://github.com/benjaminoakes/delayed-plugins-airbrake) is worth looking at?

PR #1386, opened 2016-08-23
Add a message to the load chart when the chart for a certain node is empty

### LINKED ISSUE #1380: For higher nodes, the stacked chart for gas does not show the load correctly
In [this LES](https://beta-moses.energytransitionmodel.com/testing_grounds/700), the gas-load on the lowest (end-node) level looks like:
etmoses_
Clicking a higher level, the chart freezes:
etmoses_

grdw:
Fixes #1380

### DIFF
--- a/app/assets/javascripts/graphs/d3_base_chart.js
+++ b/app/assets/javascripts/graphs/d3_base_chart.js
@@ -2,6 +2,21 @@ var D3BaseChart = (function () {
     'use strict';
 
     D3BaseChart.prototype = {
+        position: function () {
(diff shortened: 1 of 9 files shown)


antw:
screen shot 2016-08-24 at 11 09 05

antw:
Maybe we might also drop the distinction between "stacked" and "individual" in the message? If there's no load then there's no load! :grinning:

"[node] has no [carrier] load"

What do you think?

grdw:
> What do you think?

Implemented as such. Also got rid of the duplicate "'load' load"

PR #1387, opened 2016-08-24
Update synchronize profiles script

grdw:


antw:
Oh nice, I didn't realise this existed! :tada:

I think it should belong in `bin/` though, perhaps without the `.sh` extension?

```sh
git mv synchronize_profiles.sh bin/sync_profiles
```

grdw:
I think it slipped in by accident with a PR once. 😅  I was just looking at it to make it a bit more usefull than a really dumb `scp -r` command. 

> I think it should belong in bin/ though, perhaps without the .sh extension?

Yes thanks! Will do.

Issue #1375, opened 2016-08-17
[Airbrake] [Staging] undefined method `profile_required?' for nil:NilClass


dennisschoenmakers:
**Airbrake error:** [#4215](https://quintel.airbrake.io/projects/111227/groups/1754197881182994215)
**Airbrake project:** ETMoses

**Error type:** `NoMethodError`
**Error message:** `` undefined method `profile_required?' for nil:NilClass ``
**Where:** `testing_grounds#data`
**Occurred at:** Aug 17, 2016 07:57:45 UTC
**First seen at:** Aug 17, 2016 07:57:45 UTC
**Occurrences:** 1 (1 since last deploy on Aug 15, 2016 08:34:57 UTC)

**URL:** [https://beta-moses.energytransitionmodel.com/testing_grounds/271/data.json](https://beta-moses.energytransitionmodel.com/testing_grounds/271/data.json)
**File:** `[PROJECT_ROOT]/app/models/installed_technology.rb`

**Backtrace:**
```
[PROJECT_ROOT]/app/models/installed_technology.rb:289:in valid?
[PROJECT_ROOT]/app/services/testing_ground/calculator/validator.rb:14:in reject
[PROJECT_ROOT]/app/services/testing_ground/calculator/validator.rb:14:in invalid_technologies
[PROJECT_ROOT]/app/services/testing_ground/calculator/validator.rb:4:in validation_error
```

grdw:
The `households_flexibility_p2h_electricity` can't be found. It was removed in 522ceca32f4256f64d838256f765ba997f272dcf but there was no migration done for old LES's or?

grdw:
Including @antw

antw:
[There was a migration](https://github.com/quintel/etmoses/blob/master/db/migrate/20160804143207_replace_p2h_with_resistive_heaters.rb). 😕 

I'll take a look at the LES and see if I can find out why it was missed.

grdw:
Hmm. It seems to be a pretty old LES and it's yours 😅 so I don't know. Maybe the P2H is not attached to a buffer?

antw:
> Maybe the P2H is not attached to a buffer?

That's what it is, good catch! :+1:

PR #1392, opened 2016-08-25
Raise an error whenever an unknown technology is being fetched

grdw:
This is part of the problem of #1375 since that's a 'masked' error. It doesn't really tell you what the actual problem is.

### DIFF
--- a/app/models/technology.rb
+++ b/app/models/technology.rb
@@ -70,7 +70,11 @@ class Technology < ActiveHash::Base
   # Public: Retrieves the record with the matching +key+ or raises
   # ActiveRecord::RecordNotFound if no such record exists.
   def self.by_key(key)
-    key == 'generic' ? generic : where(key: key).first
+    if technology = where(key: key).first
+      technology
+    else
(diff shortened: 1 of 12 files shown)


app/models/technology.rb
```diff
@@ -70,7 +70,11 @@ def self.generic
   # Public: Retrieves the record with the matching +key+ or raises
   # ActiveRecord::RecordNotFound if no such record exists.
   def self.by_key(key)
-    key == 'generic' ? generic : where(key: key).first
+    if technology = where(key: key).first
+      technology
+    else
+      raise ActiveRecord::RecordNotFound, "can't find Technology with type #{ key }"
+    end
```
antw: This could be simplified as...

```ruby
Technology.find_by_key!(key)
```

In fact, that's *so* simple we could replace all calls to `Technology.by_key` with `find_by_key!`?

grdw: > In fact, that's so simple we could replace all calls to Technology.by_key with find_by_key!

I thought `ActiveHash::Base` objects wouldn't have such magic, but apparantly they do! 👍

PR #1396, opened 2016-08-26
Make it possible to sort dispatchable table

### LINKED ISSUE #1062: Dispatchables in heat sources need to be dragged in order

grdw:
Fixes #1062


ChaelKruip:
I have tried this locally but when I change the order, the heat load calculation does not complete pointing to this place in the code (if I am correctly reading my console debug info):
etmoses

grdw:
> I have tried this locally but when I change the order, the heat load calculation does not complete

I see why 👍 I fixed it. If you could create a new LES and try again it should work fine now.

ChaelKruip:
I have checked it locally and it works now. 

Only thing that does not seem to work is the 'save all and view heat load'. It does not seem to save changes in the order of dispatchables...

ChaelKruip:
Can this be merged @grdw?

antw:
Some feedback:

* The border above the "resize" icons is a different width to the rest of the table header:

  screen shot 2016-08-30 at 13 23 39

* The table width and height jumps slightly when grabbing and moving a row. Can that be prevented somehow? This might be hard without assigning an explicit width to each column in the table (which is not great since it breaks the responsive design).

* I'm not sold on the resize icons themselves. The up/down arrows are fine, but left/right is confusing since the rows can't be moved left or right. Changing for vertical arrows would be okay, but ideally I think this should be consistent with the ETM ("hamburger" icons and – if possible – explicit priority numbers against each row):

  screen shot 2016-08-30 at 13 25 59

  I think the *placement* of the icon – on the left of each row – is good; it doesn't matter that *that* isn't consistent with the ETM.

grdw:
> The table width and height jumps slightly when grabbing and moving a row. Can that be prevented somehow?

Hmm. It might, I have no idea how jquery-ui sortable plugin creates the 'draggable row'. But I'll give it a shot.

> "hamburger" icons

This was my first idea as an icon but `glyphicon glyphicon-menu-hamburger` icon didn't show anything 😩 . That's why I picked the cross since it means 'move' in a way. It had to do with the bootstrap version, however bumping the bootstrap version came with other complications. Ofcourse we could also just use a simple png image as an icon.

grdw:
I fixed:

- [ ] Resize icons
- [ ] Border problem
- [ ] Dragging width

ChaelKruip:
I think that the 'dragging width' is ugly but not a show-stopper at this moment. @antw do you agree that we can merge this and come back to this minor issue later?

Issue #1352, opened 2016-08-04
Cloning a LES should let you choose a new name


ChaelKruip:
Now, the same name is used which is confusing.

antw:
Since this is almost identical to the "Save a new Local Energy Solution" button on the edit page, I think upon completing this request we should also make the "Clone" button visible on private LESes (currently it shows only on those which are public).

grdw:
I was a bit to quick with that one. 

> we should also make the "Clone" button visible on private LESes 

Hmm.. should we? I don't have a really strong opinion about this matter but maybe people who work at Alliander do? What insights do you have @ChaelKruip

ChaelKruip:
> Hmm.. should we?

Yes, I think so. Only the owners (and admin's) can use this button for private LESses so it is a very logical way to go IMO.

@dennisschoenmakers @MarjoleinSchwachofer do you have an opinion?

MarjoleinSchwachofer:
It isn't directly clear to me what the difference is, so I would have to look into this.
@ThomBuijs @JeroenvdLogt: do you already have an opinion on this?

MarjoleinSchwachofer:
If I clone  my own LES, cloning works and works the same as save as, only do not have to go to click edit, go to the general page en find this button (which needs a lot of more steps and is not directly visible). Is this correct?
I tried to clone a public LES from someone else, but nothing happened. This would be a handy function  and than it would be logical to give it another name so that it can be clear which was the orginal from the other person and which one is you new LES.

grdw:
>cloning works and works the same as save as, only do not have to go to click edit, go to the general page en find this button (which needs a lot of more steps and is not directly visible). Is this correct?

Yes. This is correct.

> I tried to clone a public LES from someone else, but nothing happened

Could you be more specific? Nothing as in, the Clone button didn't work, the 'Save' button didn't bring you to the page of the cloned LES?

MarjoleinSchwachofer:
Then the button didn't work. However I tried it again today and now it worked.

ChaelKruip:
Can we close this @grdw?

grdw:
I think we can 👍

PR #1401, opened 2016-08-29
Make 'clone' function exactly the same as 'save as'

grdw:
Fixes: #1352

### DIFF
--- a/app/assets/javascripts/save_as.js
+++ b/app/assets/javascripts/save_as.js
@@ -1,4 +1,5 @@
 /*globals document,customPrompt*/
+
 $(document).on('page:change', function () {
     'use strict';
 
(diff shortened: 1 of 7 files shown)


antw:
I had to make a change to get this to work. `onSaveAs.bind(this)` was being called in the `window` context, so the callback was being bound to the `document` object and not to the button element. This meant that I would enter a name for the new LES, click "Save", and nothing would happen. It just... didn't work.

I added a namespace to the event and called `off("click.onSaveAs")` before binding; if the user moved to a different page and then hit their back button, multiple `onSaveAs` events would be bound to the same button. I think we should be using [event delegation](https://learn.jquery.com/events/event-delegation/) more.

Issue #1311, opened 2016-07-20
Round values down in the front-end, keep the original value in some part of the input


grdw:
There should be a way to split up rounded and non-rounded values in the front-end.

Something like:

````haml
%input{ value: 1.0, data: { non_rounded_value: 1.0000003 } }
````

Renders as:
````
[ 1.0 ]
````

When saving this it should look at `non_rounded_value`. However when someone changes the value of this input to '2.0' the `non_rounded_value` will just update along with it to: '2.0'. However, I can imagine this is going to get really annoying. Let's say you're really specific about your numbers and you want:

2.12345 units of something.

The front-end saves this, than when returning back to the page it rounds it down to: 

````haml
%input{ value: 2.123, data: { non_rounded_value: 2.12345 } }
````

Renders as:
````
[ 2.123 ]
````

If you want to change the last two digits... 

@ChaelKruip , what do you think? Is it likely to happen or am I just making up some once in a lifetime scenario?

ChaelKruip:
> If you want to change the last two digits...

You did successfully change those digits right? Only the front-end does not show it which is fine.

> @ChaelKruip , what do you think?

I think your approach looks fine. I'd like to hear @antw opinion too as this seems like something he might have come across.

grdw:
*Todo:*
- [ ] Never round anything (show everything raw)
- [ ] Write a little JS class that rounds 'most' inputs (sets data-raw).
- [ ] Unrounded value should be stored in `data` attribute.
- [ ] Show unrounded raw value in front-end when clicking
- [ ] Onblur should round again

PR #1405, opened 2016-08-30
Round values down in the front-end only. Keep raw values in database.

grdw:
This PR serves as to round values down to the specified rounding attribute attached to a certain `<input/>`. 

Fixes: #1311

### DIFF
--- a/app/assets/javascripts/rounder.js
+++ b/app/assets/javascripts/rounder.js
@@ -1,50 +1,76 @@
 var Rounder = (function () {
     'use strict';
 
-    function round() {
-        var rounding,
-            correction,
-            rounded,
-            value = parseFloat(this.input.val());
+    function writeRaw() {
+        this.data('raw', this.val());
+    }
 
-        if (!isNaN(value)) {
-            rounding   = value + "e+" + this.roundTo,
-            correction = "e-" + this.roundTo,
-            rounded    = +(Math.round(rounding) + correction);
+    function isValidInput(e) {
+        return e.target !== e.currentTarget &&
+            $(e.target).data('round') !== undefined;
+    }
 
-            this.input.val(rounded);
+    function onInputFocus(e) {
+        if (isValidInput(e)) {
+            var target = $(e.target);
+
+            target.val(target.data('raw'));
         }
     }
 
-    function showRawValue() {
-        var rawValue = this.input.data('raw');
+    function onInputChange(e) {
+        if (isValidInput(e)) {
+            writeRaw.call($(e.target));
+        }
 
-        this.input.val(rawValue);
+        e.stopPropagation();
     }
 
-    function writeRaw() {
-        this.input.set('raw', this.input.val());
+    function onInputBlur(e) {
+        if (isValidInput(e)) {
+            this.round.call($(e.target));
+        }
+
+        e.stopPropagation();
     }
 
     Rounder.prototype = {
+        eachInput: function (method) {
+            $(this.template).find("input[data-round]")
+                .toArray().forEach(function(input) {
+                    method.call($(input));
+                });
+        },
+
         initialize: function () {
-            writeRaw.call(this);
-            round.call(this);
+            this.eachInput(writeRaw);
+            this.eachInput(this.round);
 
-            this.input.off('focus.rounder')
-                .on('focus.rounder', showRawValue.bind(this));
+            this.template.addEventListener('focus', onInputFocus, true);
+            this.template.addEventListener('change', onInputChange, false);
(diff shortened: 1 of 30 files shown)


antw:
I like this a lot. One small niggle though: some numbers are displayed with extra trailing zeroes (two significant figures in the example below) and then change to one when the input is active:

| Inactive | Active |
| :--: | :--: |
| screen shot 2016-08-31 at 14 07 54 | screen shot 2016-08-31 at 14 08 05 |

It would be nice if these were consistent: if the inactive number is showing two significant figures then the active version should also show at least two figures. A better alternative might be for the inactive (rounded) version to be rounded to one significant figure *in this case* (`10.50` → `10.5`), since the extra zero doesn't serve any purpose.

grdw:
> A better alternative might be for the inactive (rounded) version to be rounded to one significant figure in this case

I agree however I showed it to @ChaelKruip like that and he disagreed due to the lack of consistency. This is only showing the significant values:

screen shot 2016-08-31 at 17 00 42

Some of these numbers have (`,0` or `,045435`) which all boils down to nothing.

ChaelKruip:
> I agree however I showed it to @ChaelKruip like that and he disagreed due to the lack of consistency. 

I can't remember disagreeing (although I might have just to be difficult) but am now totally fine with not showing insignificant zero's. Everybody happy?

antw:
Functionality looks good, but I'm seeing three things which are problematic:

* If you add a new technology to the form, the values aren't rounded (to reproduce: 1. Visit "Edit LES" -> "Technologies", 2. Add a battery to the form).

  screen shot 2016-09-01 at 14 27 18

* There are inputs with `data-round` attributes which shouldn't have them:

  * `type`
  * `node`
  * `profile_key`
  * `components`
  * `associates`
  * `composite_index`
  * ...

* Performance when setting up the form is not good. A LES with 24 technologies spends 260ms just running `new Rounder().initialize`. See: [Before](https://cloud.githubusercontent.com/assets/4383/18169142/aec679fe-7050-11e6-8825-ca8db4aa1f2e.png) and [after](https://cloud.githubusercontent.com/assets/4383/18169165/c1021bf0-7050-11e6-85fe-fa60e065cb78.png) (that `each` consists of [a load of calls](https://github.com/quintel/etmoses/pull/1405/files#diff-9538b5cc203645fdd14c863f7f10e17cR79) to `new Rounder().initialize`). A LES with 100+ techs [takes more than a second to set up Rounders](https://cloud.githubusercontent.com/assets/4383/18169503/48df14be-7052-11e6-8032-88b0ca8f7932.png).  

  As far as I can tell, the main culprit is the time it takes to call `off()` and `on()` to set up events on all of the `input` elements. I created [a version avoids this using event delegation](https://gist.github.com/antw/5efad4b1821622699b2b8225699760e7) (only one event is created on the parent `form` element). However this doesn't work correctly: whenever you blur one of the `inputs` the value is reset to the original value.

  I'm guessing there's an onChange on the `input`, which parses the form to JSON and that's getting called before the delegated `writeRaw` and `round` events.

  I'm not sure what to do about this. I think taking a second to set up Rounders is rather excessive, but the alternative is to change all events on this form to use delegation. Switching to delegation would be a good idea since performance of the form initializer is pretty bad, but I expect that's quite a large task. Delegation also partially fixes problem 1, since you don't have to bind focus/blur/change events for the new fields (but the initial set-up is still required).

  A stop-gap solution might be to wait until the visitor `focus`es a field before setting up the `change` and `blur` events.

  ```js
  input.on('focus', function() {
      if (this is the first time the input is focused) {
          // set up change event
          // set up blur event
      }
  
      round(input);
  }
  ```

grdw:
Thanks for all the input @antw! Lots of it is fixed in the optimalizations commit however I still need to check if it performs well in older browsers. Also I think I can make it just even faster but I need to check. 👍 

This is the current profile:
screen shot 2016-09-02 at 12 42 46

*Todo still:*
- [ ] check slider of congestion
- [ ] check Safari
- [ ] chekc IE

app/assets/javascripts/rounder.js
```diff
@@ -0,0 +1,77 @@
...
+                .toArray().forEach(function(input) {
+                    method.call($(input));
+                });
+        },
+
+        initialize: function () {
+            this.eachInput(writeRaw);
+            this.eachInput(this.round);
+
+            this.template.addEventListener('focus', onInputFocus, true);
```
antw: Nice! I didn't know about the `useCapture` param; that's useful! :+1:

antw:
Great changes @grdw. I can't even find the Rounder in the timeline anymore, so that's a huge improvement. I'll merge this now...

Issue #1090, opened 2016-06-15
The business case table should show a breakdown of the number in each cell on mouse-over


ChaelKruip:
Right now it is very difficult to see what the origin of outcomes of the cost calculation are.

ChaelKruip:
I think that we should support mouse-over break down of numbers for the business case summary shown in the results section:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/17971773/81c1faf6-6adc-11e6-8064-ac44a770115c.png)
Perhaps we can also support mouse-over break down for the matrix in the edit section.

ChaelKruip:
I am still unsure to what level of detail we break down the costs. My first guess would be to identify:

* Depreciation costs of [carrier] infra
* O&M costs of [carrier] infra
* Depreciation costs of district heat sources
* O&M costs of district heat sources
* Result of every relevant row (transaction) of the Market Model

The results could perhaps be displayed as follows:

```
* Depreciation costs of gas infra              = 12345.6 EUR
* O&M costs of gas infra                       = 78.9 EUR
* Payed by [payer] for [measure] with [tariff] = 1234.5 EUR
------------------------------------------------------------- + 
 TOTAL                                         = xxxx EUR
```

ChaelKruip:
@grdw @dennisschoenmakers @antw @MarjoleinSchwachofer @ThomBuijs what do you think would be useful?

MarjoleinSchwachofer:
I think this indeed would be a useful breakdown.

Another nice to have would be if you could export the businesscase and breakdown in an Excel or something and not only see them in the mouseover. So that you can copy them easily.

grdw:
![screen shot 2016-09-01 at 10 07 20](https://cloud.githubusercontent.com/assets/2676542/18160119/1aa0add6-702c-11e6-8a2b-b7820b06692f.png)

Right now it shows the breakdown of the total numbers. It does not yet know if the calculated number came from O&M costs or were depreciation costs etc.

PR #1408, opened 2016-09-02
Show basic breakdown of Business case costs

grdw:
Fixes: #1090

### DIFF
--- a/app/assets/javascripts/business_case.js
+++ b/app/assets/javascripts/business_case.js
@@ -1,0 +1,13 @@
+var BusinessCase = (function () {
+    'use strict';
+
+    return {
+        formatCell: function () {
+            if ($(this).text().replace(/\s/g, '') === '') {
+                $(this).html("--");
+            } else {
+                $(this).formatCurrency();
+            }
+        }
+    }
+}());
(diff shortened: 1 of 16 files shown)


antw:
* Is it intended for the individual market model rows to *not* be included in the breakdown?

  screen shot 2016-09-05 at 16 21 26

  I think that €30k "Customer" line is actually a market model transaction between "Customer" and "Supplier" for "kWh of electricity consumed". I gathered [from this post](https://github.com/quintel/etmoses/issues/1090#issuecomment-242415322) that each individual market model row would be included in the summary, but perhaps the plan has changed since then?

  I'm guessing it's not possible to show that information without recomputing the market model (or storing the outcome of each individual transaction)...

* The numbers should use the same formatting as the rest of the app: `€30.811,70` → `€30,811.70`. `number_to_currency` appears to be formatting the number correctly in the BusinessCaseHelper, but it's getting changed somewhere else (JS?).

* Clicking on a number in the business case table reveals that the `span.static` has a `:focus` style, which I don't think should be there.

  | Chrome | Firefox |
  | :--: | :--: |
  | screen shot 2016-09-05 at 16 33 06 | screen shot 2016-09-05 at 16 33 03 |

grdw:
> Is it intended for the individual market model rows to not be included in the breakdown?

For now, yes.

> The numbers should use the same formatting

Checked!

> Clicking on a number in the business case table reveals that the span.static has a :focus style, which I don't think should be there.

Also fixed that 👍

Issue #1422, opened 2016-09-12
Clear business case jobs when market model is edited


grdw:
Pseudo code:

````ruby
...
before_filter :clear_business_case_jobs, only: :update
...

private

def clear_business_case_jobs
  testing_ground_ids = TestingGround.where(market_model_id: @market_model.id).pluck(:id)
  BusinessCase.where(testing_ground_id: testing_ground_ids).each(&:clear_job!)
end
````

PR #1430, opened 2016-09-16
Recalculate a user's business case when the market model is updated

grdw:
Fixes: #1422

### DIFF
--- a/app/controllers/market_models_controller.rb
+++ b/app/controllers/market_models_controller.rb
@@ -33,7 +33,7 @@ class MarketModelsController < ResourceController
   end
 
   def update
-    if @market_model.update(market_model_params)
+    if MarketModel::Update.update(@market_model, market_model_params)
       redirect_to market_model_path(@market_model)
     else
       render :edit
--- a/spec/controllers/market_models_controller_spec.rb
+++ b/spec/controllers/market_models_controller_spec.rb
@@ -34,4 +34,39 @@ RSpec.describe MarketModelsController do
       expect(response).to redirect_to(new_user_session_path)
     end
   end
+
+  describe "#update" do
+    let(:user) { FactoryGirl.create(:user) }
+    let(:market_model) { FactoryGirl.create(:market_model, user: user) }
+    let(:business_cases) {
+      BusinessCase.where(testing_ground: TestingGround.where(market_model: market_model))
+    }
+
+    let(:control_group) {
+      FactoryGirl.create(:business_case, job_id: 2)
+    }
+
+    before do
+      sign_in(:user, user)
+
+      3.times do
+        testing_ground = FactoryGirl.create(:testing_ground,
+          market_model: market_model)
+
+        FactoryGirl.create(:business_case,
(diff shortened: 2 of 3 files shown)


app/controllers/market_models_controller.rb
```diff
@@ -56,6 +57,11 @@ def destroy
 
   private
 
+  def clear_business_case_jobs
```
antw: This would be a good candidate for a service object, e.g.:
```ruby
if MarketModel::Update.run(market_model, market_model_params)
  redirect_to ... # success
else
  render :edit # validation errors
end
```

(like [TestingGround::SaveAs](https://github.com/quintel/etmoses/blob/4c45e7383e74d9d8151384981ab04fa8576003f9/app/services/testing_ground/save_as.rb)).

spec/controllers/market_models_controller_spec.rb
```diff
@@ -34,4 +34,28 @@
...
+        FactoryGirl.create(:business_case,
+                           testing_ground: testing_ground, job_id: 1)
+      end
+    end
+
+    it 'should recalculate the business case for all LESs' do
+      patch :update, id: market_model.id, market_model: {
+        name: "test", public: true, interactions: [] }
+
+      expect(BusinessCase.pluck(:job_id).compact).to be_empty
```
antw: An explicit check for `expect(...).to eq([nil, nil, nil])` would be better. What if `patch :update` accidentally *deleted* all the business cases? This test would still pass.

It would also be worthwhile to assert that business cases which use a *different* market model are unaffected by the action. If the action instead performed `BusinessCase.update_all(job_id: nil)` the tests would still pass.


```ruby
let(:user) { FactoryGirl.create(:user) }
let(:market_model) { FactoryGirl.create(:market_model, user: user) }
let(:other_market_model) { FactoryGirl.create(:market_model) }

let(:les1) { FactoryGirl.create(:testing_ground, market_model: market_model) }
let(:les2) { FactoryGirl.create(:testing_ground, market_model: other_market_model) }

let(:bc1) { FactoryGirl.create(:business_case, testing_ground: les1, job_id: 1) }
let(:bc2) { FactoryGirl.create(:business_case, testing_ground: les1, job_id: 2) }
let(:bc3) { FactoryGirl.create(:business_case, testing_ground: les2, job_id: 3) }

let(:perform_action) do
  patch :update, id: market_model.id,
    market_model: { name: "test", public: true, interactions: [] }
end

# ...

it 'invalidates job_id on all associated business cases' do
  expect { perform_action }
    .to change { [bc1, bc2].map { |bc| bc.reload.job_id } }
    .from([1, 2]).to([nil, nil])
end

it 'does not invalidate job_id on unrelated business cases' do
  expect { perform_action }.not_to change { bc3.reload.job_id }.from(3)
end
```

grdw:
I added a control group in the specs to validate that no other business cases are being reset. I also moved the relevant code to a separate service.

antw:
LGTM; thanks! :+1:

Issue #1458, opened 2016-09-28
When adding a new node the name is blank


grdw:
Produce a useful error message when the user hits the save button without changing the new. Prefereably the name input will be a reddish colour + the node will turn red.

PR #1462, opened 2016-09-28
Topology/save validation

grdw:
Fixes: #1458

### DIFF
--- a/app/assets/javascripts/save_all.js
+++ b/app/assets/javascripts/save_all.js
@@ -26,6 +26,19 @@ var SaveAll = (function () {
         }
     }
 
+    function submit() {
+        $(this).find("input[type=submit]").addClass("disabled");
+        $(this).find("span.wait").removeClass("hidden");
+    }
+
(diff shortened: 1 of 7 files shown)


antw:
I know it's a work-in-progress (which I guess is why you want this merging into the "new-topology-editor" branch?)... One small thing I noticed: on the LES topology page – `/testing_grounds/:id/edit` – if there's a validation error on a node the "Save" button is still disabled and the spinner appears:

screen shot 2016-09-29 at 11 01 38

I do like what I see though, the new editor is looking great. :+1:

grdw:
> which I guess is why you want this merging into the "new-topology-editor" branch

Yes. It's not yet in a stable state.

> if there's a validation error on a node the "Save" button is still disabled and the spinner appears:

Thanks for checking.

grdw:
> if there's a validation error on a node the "Save" button is still disabled and the spinner appears:

I fixed that problem 👍

Issue #1470, opened 2016-10-03
The producer should not have depreciation costs as income


ChaelKruip:
![screen shot 2016-10-03 at 14 50 56](https://cloud.githubusercontent.com/assets/1303760/19037851/d2f98e4c-8978-11e6-83b8-7c24cf327552.png)

Happens in [this LES](https://beta-moses.energytransitionmodel.com/testing_grounds/736/business_cases/497)

PR #1471, opened 2016-10-03
Make sure the breakdown skips over the appropriate stakeholder

grdw:
Fixes: #1470

### DIFF
--- a/app/services/finance/business_case_summary/summarizer.rb
+++ b/app/services/finance/business_case_summary/summarizer.rb
@@ -4,16 +4,15 @@ module Finance
       include Breakdown
 
       def summarize_rule(financial_rule, index)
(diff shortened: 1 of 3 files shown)


app/services/finance/business_case_summary/summarizer.rb
```diff
@@ -27,8 +27,12 @@ def stakeholder
         @financial_rule.keys.first
       end
 
+      def incoming
```
antw: I think it's confusing to have an instance variable called `@incoming` and an instance method also called `incoming`. Could the latter be renamed to something more meaningful? `incoming_from_rule` or something like that?

grdw: I agree and changed it. `incoming_from_rule` sounds fine by me. 👍

Issue #1434, opened 2016-09-19
New topology editor overview issue


grdw:
- For the interface elements -> #1432
- D3 Adding/removing/positioning nodes -> #1438
- For the styling/looks/feel -> #1433
- For technical -> 
   - Technical design (Javascript) -> #1435

Issue #1432, opened 2016-09-19
Inteface elements of new topology interface


grdw:
**Adding a new node**
We could also when hovering/focussing over/on a node add another bulb/'add/+' button (like here) to be able to drag a new node out of. This way moving a node can be just done by dragging without having to press Ctrl/Cmd. Also the 'add/+' button will be able to take two event handlers:

```javascript
addButton
    .on('click', function () {
        // Move the new node directly under the node
    })
    .on('dragend', function () {
        // Set the node on the position where the user left it.
    })
```

What if someone clicks the addButton 3x?

```
First time:       | Second time:    | Third time:
O +                  O +               O +
|                   / \              / | \
O                  O   O            O  O  O

```
It should automatically space out the nodes under the 'clicked' node. What if someone accidentally drags his node over his own node? Well, than he can drag it back.

**Moving a node**
Moving a node I would do by hitting the Ctrl/Cmd button (depending on which idea is suitable for adding a node) + dragging the node. Clicking on a node will put the node in 'focus' in essence giving it a different glow + showing the information form.

**Removing a node**
First off the most obvious choice here is to be able to click a node and hit the DEL/Backspace key. This is similar to the example above. This removes all nodes and children under that node (I think that when you have children under a node it should alert with "Do you really want to remove your children?" or something a bit more parent-friendly. 

---

Interface iteration 1:

![screen shot 2016-09-19 at 10 34 52](https://cloud.githubusercontent.com/assets/2676542/18626567/222969ac-7e56-11e6-91f8-b001223abc81.png)

grdw:
![screen shot 2016-09-28 at 11 54 00](https://cloud.githubusercontent.com/assets/2676542/18909020/43c1c870-8572-11e6-875a-9097bbfda651.png)

Overview of interface.

You are able to add, update and remove nodes.

Issue #1412, opened 2016-09-06
[Airbrake] [Production] Validation failed: Technology profile includes a connection to missing node "Eengezin 55a", Technology profile includes a connection to missing node "Eengezin 35a", Technology profile includes a connection to missing node "Eenge...


dennisschoenmakers:
**Airbrake error:** [#4956](https://quintel.airbrake.io/projects/111227/groups/1768656416306344956)
**Airbrake project:** ETMoses

**Error type:** `ActiveRecord::RecordInvalid`
**Error message:** `` Validation failed: Technology profile includes a connection to missing node "Eengezin 55a", Technology profile includes a connection to missing node "Eengezin 35a", Technology profile includes a connection to missing node "Eengezin 45a", Technology profile includes a connection to missing node "Eengezin 36a", Technology profile includes a connection to missing node "Eengezin 8b", Technology profile includes a connection to missing node "Eengezin 35b", Technology profile includes a connection to  ``
**Where:** `testing_grounds#save_as`
**Occurred at:** Sep 06, 2016 06:44:34 UTC
**First seen at:** Sep 06, 2016 06:44:34 UTC
**Occurrences:** 1 (1 since last deploy on Jul 18, 2016 11:54:39 UTC)

**URL:** [https://moses.energytransitionmodel.com/testing_grounds/518/save_as](https://moses.energytransitionmodel.com/testing_grounds/518/save_as)
**File:** `[GEM_ROOT]/gems/activerecord-4.1.8/lib/active_record/validations.rb`

**Backtrace:**
```
[GEM_ROOT]/gems/activerecord-4.1.8/lib/active_record/validations.rb:57:in save!
[GEM_ROOT]/gems/activerecord-4.1.8/lib/active_record/attribute_methods/dirty.rb:29:in save!
[GEM_ROOT]/gems/activerecord-4.1.8/lib/active_record/transactions.rb:273:in block in save!
[GEM_ROOT]/gems/activerecord-4.1.8/lib/active_record/transactions.rb:329:in block in with_transaction_returning_status
```

grdw:
Duplicate of #1399

PR #1484, opened 2016-10-06
New topology editor

grdw:
Initial graphic topology editor.

- [x] JSLinting

---

Closes: #1412 
Closes: #1433 
Closes: #1432 
Closes: #1434


grdw:
> When adding a new node to an existing topology, the "Name" field remains disabled. Naturally the validation fails since I can't enter a name for the node.

Hmmm. yes this was to prevent people from renaming existing nodes. Ofcourse this fix is a little patchy. I'll re-enable the field and figure out something better.

> If I add a node and then change my mind, there's no way to remove the node except to refresh the browser and start over. 

You can hit the `delete` key 👍 


---

I manage to fix most of these points. Some of them are still open though like the dragging and  "it should not be possible to add child nodes to an endpoint which already has technologies attached.".

app/assets/javascripts/topology/graph/editor_graph.js
```diff
@@ -0,0 +1,129 @@
...
+                .attr("class", "add-button")
+                .attr('d', plusPath(addSize))
+                .attr('stroke', 'green')
+                .attr('stroke-width', 3)
+                .attr('fill', 'none')
+                .attr('transform', 'translate(22, 0)')
+                .on('click', function (d) {
+                    window.TopologyEditor.graphEditor.addNode(d);
+                })
+                .on("mouseover", function () {
```
grdw: Hmm I think I stumbled upon the same issue as you:

> makes the button frustratingly difficult to click

But didn't think it were my clicking skills and went for this farfetched solution. Yes the `cursor: pointer`  in the SASS just works. 😅

antw:
> I managed to fix most of these points.

I've got more. Sorry. :cry:

* The topology preview on the "Import" page is missing node(s) sometimes. This is topology ID:529 ("Lochem Weidemolen MSR Thom") from the production server:

  | Production | `new-topology-editor` |
  | :---: | :---: |
  | screen shot 2016-10-11 at 13 09 55 | screen shot 2016-10-11 at 13 10 50 |

* > You can hit the delete key 👍

  Thanks. IMO when you can *add* a node with a button it's not obvious or intuitive that pressing the delete key is how to remove one. A `-` button would be more consistent. Interested in hearing what others think about this...

* When editing a topology, it seems I can only delete nodes which existed when the page loaded. If I add a new node, the delete key doesn't do anything.

* I created a new LES using the default topology, chose to edit the LES and selected the topology tab. Clicking a node and tapping "Delete" locked up the whole tab. 😕  (Chrome 53.0.2785.143 and Firefox 48.0.2). It happens consistently when I try to delete the highlighted nodes:

  screen shot 2016-10-11 at 13 20 49

  I can delete the parent "HV-MV Trafo" node just fine though. This also seems to happen (sometimes) when editing one of the attributes for these nodes.

* You still have to click precisely on the green bars of the `+` to add a node. Is this a limitation of SVG? Perhaps another (larger) element can be placed behind the button to increase the clickable area?

  For example, this small ETModel button has a larger clickable area than is visible to the user:

  screen shot 2016-10-11 at 13 57 47

* The `+` button looks kind of fat now. I guess this was because it's not possible to keep the stroke width at 3px while keeping the button vertically centered with the node?

  Perhaps is makes sense for the button to be off-center, to more clearly show that the button adds a child instead of a sibling?

  screen shot 2016-10-11 at 13 49 40

  Another alternative (although more complex) would be for a clicked node to transform into "Add" and "Remove" buttons:

  ```
     _|_            ___|___
    (   )    ->    ( - | + )
     ‾|‾            ‾‾‾|‾‾‾
  ```

ChaelKruip:
> Another alternative (although more complex) would be for a clicked node to transform into "Add" and "Remove" buttons:

I like that!

grdw:
- [ ] "The topology preview on the "Import" page is missing node(s) sometimes."
- [ ] "When editing a topology, it seems I can only delete nodes which existed when the page loaded"
- [ ] "I created a new LES using the default topology, chose to edit the LES and selected the topology tab. Clicking a node and tapping "Delete" locked up the whole tab."
- [ ] "You still have to click precisely on the green bars of the + to add a node"
- [ ] "Perhaps is makes sense for the button to be off-center, to more clearly show that the button adds a child instead of a sibling?"

---

> Another alternative (although more complex) would be for a clicked node to transform into "Add" and "Remove" buttons:

This is going to be rather ugly. I'd rather have some sort of pop-up instead when clicking the node which presents a tiny little interface where you can choose between "add" or "remove". Especially in light of the advanced features which one day might be implemented.

grdw:
@antw I think this is good to go now 👍  I manage to fix the dragging node problem so I hope all is fine.

antw:
> so I hope all is fine.

Great job @grdw, it's looking a whole lot better now. :+1:

Everything looks good with two exceptions. On the LES topology page (testing_grounds/:id/edit):

* "Save and view LES" button doesn't seem to save the changes.
* There appears to be no "add child" / "delete node" popup on the LES topology page; was that intentional?

grdw:
> Save a;; and view LES" button doesn't seem to save the changes.

I shall take a look.

> There appears to be no "add child" / "delete node" popup on the LES topology page; was that intentional?

Yes. It's a bit weird to add a node or remove a node when you're in the LES topology page (you already chose a topology to begin with). Also you can't add an endpoint for that will result in a validation error.

When we're finally going towards #1488 than it would maybe be less of a problem to re-enable the interface for "Adding and removing a node" since every topology is a duplicate anyway and the `testing_ground -> technology_profile` and `topology -> graph` columns might be merged into one column.

antw:
> Yes. It's a bit weird to add a node or remove a node when you're in the LES topology page (you already chose a topology to begin with). Also you can't add an endpoint for that will result in a validation error.

I'm not so sure about that. It's an error to delete an endpoint (which might have technologies attached), but adding new ones should be okay.

For example:

| Before | After |
| --- | --- |
| screen shot 2016-10-14 at 12 55 54 | screen shot 2016-10-14 at 12 56 23 |

Technically you *can* delete an endpoint provided you add a new one with the same name elsewhere in the topology.

This might be a problem if a user wants to create a custom version of one of their own topologies – [such as this](https://moses.energytransitionmodel.com/topologies/498). Since we don't have a "Save as / clone" button for topologies, it means they'd have to create the topology from scratch every time they want to make a change to the structure.

I'm okay with this if #1488 is something we plan to do very soon. @ChaelKruip?

grdw:
> I'm not so sure about that. It's an error to delete an endpoint (which might have technologies attached), but adding new ones should be okay.

It's not what I currently experience on beta. 

screen shot 2016-10-14 at 14 14 03

If I add a new endpoint called Households 4 I'm getting the 'deleted' error. Which is a bit odd, since I added one. 😆

antw:
[This is why](https://github.com/quintel/etmoses/blob/8c5d49c4f0e86d892edf7f895480af8077fb1d75/app/models/topology.rb#L115). That's an odd validation. 😆 

I'm a little worried that the inability to add nodes on the LES page might push visitors into making their changes on the "edit topology" page (`topologies/:id/edit`) instead. A change there doesn't result in a clone and might break other LESes which also use it.

I still think this is okay as long as #1488 is something we do soon.

grdw:
> This is why. That's an odd validation.

Man... I know I wrote that. Just changing the `!=` into  a `<` makes it actually do what the error message is describing. 😅 

I'm also ok with enabling the adding and removing but I purely judged by the inability of adding and removing a node by the error I got and therefor disabled the interface.

I do still want to disable renaming the nodes for each node which has technologies attached to it for that will currently result in errors.

ChaelKruip:
> I'm okay with this if #1488 is something we plan to do very soon. @ChaelKruip?

I think we should! @grdw can you pick that up after this?

antw:
Looks good now. Merged it! :tada: :cake:

Issue #1447, opened 2016-09-20
Remove opacties of colors


grdw:
It becomes a little confusing:

<img src="http://g.recordit.co/OibEg0T9qk.gif"/>

Issue #1087, opened 2016-06-14
Convert LoadChartsSettings to something like OutputElementSeries in ETModel


grdw:
I'm noticing that the [LoadChartsSettings](https://github.com/quintel/etmoses/blob/master/app/assets/javascripts/les/topology_load_charts/load_charts_settings.js) file is getting really fat. Also there are now multiple ways of drawing a chart which I would like to reduce to only a single way.

I actually like how that's done in ETModel so maybe we could (re)-use those parts for ETMoses?

PR #1490, opened 2016-10-11
Clean up settings

grdw:
~~First merge #1484~~

Fixes: #1447 
Fixes: #1087


antw:
The "Gas Load" → "Load on the gas network" chart won't load for me. This is because there is no ChartSetting for `gas`. The chart itself shouldn't actually need a `gas` setting though (it uses `gas_local`, `gas_four`, etc). Perhaps because the chart data has `key: "gas"`?

```js
{
  key: "gas"
  name: "Gas load chart",
  values: [ /* ... */ ]
}
```

* TypeError: Cannot read property 'axisLabel' of undefined: [d3_load_chart.js, line 282 ](https://github.com/quintel/etmoses/blob/a261a88b3a468ec85af078247624b44329dd6334/app/assets/javascripts/graphs/d3_load_chart.js#L280-L281)

--

Could `ChartSettings.forChart` perhaps also throw an error if the setting doesn't exist? For example:

```js
forChart: function (chartKey) {
  if (!this.data.hasOwnProperty(chartKey)) {
    throw new Error('No such chart setting: ' + chartKey);
  }

  return this.data[chartKey];
}
```

grdw:
@antw I think all is good 👍

antw:
> I think all is good 👍

Not for me. 😧 

* The heat chart seems to be missing settings for some series, which is causing them to appear without a label and black as the colour. Note also that the pop-over line shows "0" as the value for every series; this happens on all charts.

  screen shot 2016-10-19 at 14 07 56

* The auto-scaling of the chart units is *slightly* wrong.

  screen shot 2016-10-19 at 14 07 06

antw:
> The heat chart seems to be missing settings for some series, which is causing them to appear without a label and black as the colour.

I can shed a bit more light on this. The settings are not missing, but just don't have all the values needed for the chart. The affected series are the central heat producers (households collective CHP biogas, households collective geothermal, etc).

For these series, ChartSettings.forChart is returning something like:

```js
{
  type: "households_collective_chp_network_gas",
  visible: true
}
```

... but for other series, is returning:

```js
{
  areaColor: "#9CC7F9",
  axisLabel : "kW",
  color : "#9CC7F9",
  id: 33,
  name: "Hot water demand",
  type: "water_heating",
  visible : true
}
```

grdw:
> The pop-over line shows "0" as the value for every series; this happens on all charts.

That's because of:

> The auto-scaling of the chart units is slightly wrong.

I was a bit quick to call it 'okido'. I need to test a bit more thoroughly I guess. I'm sorry.

I know about the colouring problem. I thought I had all of them.

antw:
LGTM! :+1:

Issue #1474, opened 2016-10-04
[Airbrake] [Production] Validation failed: Name is too long (maximum is 100 characters)


dennisschoenmakers:
**Airbrake error:** [#1750](https://quintel.airbrake.io/projects/111227/groups/1789160003738251750)
**Airbrake project:** ETMoses

**Error type:** `ActiveRecord::RecordInvalid`
**Error message:** `` Validation failed: Name is too long (maximum is 100 characters) ``
**Where:** `testing_grounds#save_as`
**Occurred at:** Oct 04, 2016 13:41:26 UTC
**First seen at:** Oct 04, 2016 13:41:26 UTC
**Occurrences:** 1 (1 since last deploy on Sep 29, 2016 12:50:22 UTC)

**URL:** [https://moses.energytransitionmodel.com/testing_grounds/578/save_as](https://moses.energytransitionmodel.com/testing_grounds/578/save_as)
**File:** `[GEM_ROOT]/gems/activerecord-4.1.16/lib/active_record/validations.rb`

**Backtrace:**
```
[GEM_ROOT]/gems/activerecord-4.1.16/lib/active_record/validations.rb:57:in save!
[GEM_ROOT]/gems/activerecord-4.1.16/lib/active_record/attribute_methods/dirty.rb:29:in save!
[GEM_ROOT]/gems/activerecord-4.1.16/lib/active_record/transactions.rb:273:in block in save!
[GEM_ROOT]/gems/activerecord-4.1.16/lib/active_record/transactions.rb:329:in block in with_transaction_returning_status
```

grdw:
A validation shouldn't trigger an error ...

dennisschoenmakers:
*Comment from Airbrake:* Resolved [error](https://quintel.airbrake.io/projects/111227/groups/1789160003738251750) reoccurred.

grdw:
This has to be the `production` environment triggering this error.

PR #1493, opened 2016-10-11
Validate name input before submitting it to the server

grdw:
Fixes: #1474


app/assets/javascripts/save_as.js
```diff
@@ -1,40 +1,77 @@
...
             submitSaveAsForm.call(this);
         } else {
-            customPrompt(
-                "Please set a name for your new LES",
-                nameInput.val(),
-                promptCallback.bind(this)
+            this.customPrompt = CustomPrompt.prompt(
+                "Please set a new name for your LES",
+                this.nameInput.val(),
+                callback.bind(this)
```
antw: I think it is possible to have `callback` take an explicit parameter. Instead of:

```js
function callback(nameInput) {
  // A wild "this" appears!
}
```

... you might try ...

```js
function callback(saveAsForm, nameInput) {
  // ...
}
```

... and then replace `callback.bind(this)` with `callback.bind(null, this)` (which creates a [partially-applied function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind#Partially_applied_functions)). The callback function will work identically when triggered by `CustomPrompt`, but has the benefit of being clearer about what argument its needs.

app/assets/stylesheets/general_style.css.sass
```diff
@@ -72,6 +72,10 @@ i.last-updated
   input.prompt-message
     margin: 5px 0
 
+input.invalid
+  border-color: red
+  background-color: #ffeeee
+
```
antw: Might want to reuse the [built-in Bootstrap CSS classes for this](http://getbootstrap.com/css/#forms-control-validation)?

grdw:
> How about catching the exception in the controller and returning JSON to the browser with the list of errors, and then presenting them to the user on the "Save as" form?

This makes more sense :)

grdw:
> Name must not include an obscure sci-fi reference

😆

app/assets/javascripts/save_as.js
grdw: I fixed it a little bit different. I managed to just get rid of all arguments 😆 If you could have a look 👍

app/assets/javascripts/save_as.js
```diff
@@ -1,40 +1,77 @@
...
-        form.submit();
+        this.form
+            .attr("action", this.formAction)
+            .submit();
     }
 
-    function promptCallback(newName) {
-        nameInput.val(newName);
+    function isValid(value) {
+        return (value.length < 100 && value.length > 0);
```
grdw: I removed this validation and instead use the server to validate the input as suggested.

app/services/testing_ground/save_as.rb
```diff
@@ -18,14 +18,16 @@ def initialize(original, name, user = nil)
 
   def run
     duplicate = @original.dup
+    duplicate.name = @name
+    duplicate.user = @user
 
-    TestingGround.transaction do
-      duplicate.name = @name
-      duplicate.user = @user
+    if duplicate.valid?
```
antw: The whitespace on these lines is a bit wonky; there are a couple of blank lines you probably didn't mean to commit:

```ruby
if duplicate.valid?
  TestingGround.transaction do

    duplicate.save!

    duplicate_associations!(duplicate)
  end
end
```

It's not really necessary to run `#valid?` here since `save!` checks the object validity already (these three lines effectively run `duplicate#valid?` twice). It also possible that one of the duplicated associations raises an error due to being invalid, and `duplicate#valid?` won't catch that.

An alternative would be to skip the `valid?`, let the exception get raised, and catch it in the controller like so:

```ruby
def save_as
  @testing_ground = TestingGround::SaveAs.run(
    @testing_ground, testing_ground_params[:name], current_user
  )

  render json: { redirect: testing_ground_path(@testing_ground) }
catch ActiveRecord::RecordInvalid
  render json: { errors: @testing_ground.errors }, status: 422
end
```

grdw: > and catch it in the controller like so:

This was my first approach. But I did it like this:

````ruby
def save_as
  begin
    @testing_ground = TestingGround::SaveAs.run(
      @testing_ground, testing_ground_params[:name], current_user
    )

  render json: { redirect: testing_ground_path(@testing_ground) }
rescue ActiveRecord::RecordInvalid
  render json: { errors: @testing_ground.errors }, status: 422
end
````

However this caused the `@testing_ground` to point to the old testing ground and not towards the duplicate. Maybe `catch` prevents this behavior though.

antw: > Maybe catch prevents this behavior though.

My bad. I meant `rescue`; catch is something different altogether. Too much time writing JavaScript lately... 😆

antw: Apparently the record is available on the `RecordInvalid` exception:

```ruby
def save_as
  begin
    @testing_ground = TestingGround::SaveAs.run(
      @testing_ground, testing_ground_params[:name], current_user
    )

  render json: { redirect: testing_ground_path(@testing_ground) }
rescue ActiveRecord::RecordInvalid => ex
  render json: { errors: ex.record.errors }, status: 422
end

antw:
LGTM! Thanks for the changes @grdw! :tada:

Issue #847, opened 2016-03-10
It should be more apparent if a P2H unit is placed


ChaelKruip:
![etmoses](https://cloud.githubusercontent.com/assets/1303760/13672808/96ca6706-e6d6-11e5-8add-ab81aa63d263.png)

Now nothing happens on the screen...

grdw:
It does strike me as annoying as well. So maybe we could, when somebody adds a technology that belongs to a buffer, automatically do an animated scroll to the added technology?

grdw:
However, before implementing anything like that I'm waiting for the #1322 PR to be merged. Since in that PR there are a lot of changes regarding the profile matrix editor.

grdw:
I discussed with @ChaelKruip and found out that scrolling towards the newly added technology is acutally quite nice. 😄 However, for this to work even better we thought of adding a "focus" style. I'm playing a bit with what looks 'nice'. I used a glowing border first but that didn't look so pretty:

![screen shot 2016-10-12 at 16 15 18](https://cloud.githubusercontent.com/assets/2676542/19313533/18c88e26-9097-11e6-9c2f-caf77ab3bce1.png)

Than I added a bar to the left:

![screen shot 2016-10-12 at 16 13 33](https://cloud.githubusercontent.com/assets/2676542/19313510/0725b126-9097-11e6-9aac-8c731478822f.png)

Maybe it should be blue aswell?

![screen shot 2016-10-12 at 16 18 59](https://cloud.githubusercontent.com/assets/2676542/19313652/99fe8f04-9097-11e6-8ed4-a96329aae9d2.png)

In an overview:

![screen shot 2016-10-12 at 16 19 19](https://cloud.githubusercontent.com/assets/2676542/19313672/a8111fb2-9097-11e6-93a0-6c6c7d0b0489.png)

Any preferences or suggestions?

ChaelKruip:
> Maybe it should be blue aswell?

I prefer the blue over the green and glowing but that's a personal taste I guess 🙂

grdw:
@ChaelKruip do you mean glowing - as in - all around glowing? So option 1?

ChaelKruip:
> @ChaelKruip do you mean glowing - as in - all around glowing? So option 1?

No, I mean that the blue bar has my preference.

PR #1496, opened 2016-10-12
Scroll towards the newly added buffer child

grdw:
+ Adding a focus style to whichever technology is last clicked upon

Fixes: #847


antw:
Nice! :+1:

Issue #1499, opened 2016-10-17
Wanted, clone option in "edit topology"


DorinevanderVlies:
When working on a topology it can be useful to use an existing topology to start with. Currently there is no `clone` option when editing a topology. 

Previously this was less necessary as the YAML code of an existing topology could be copied to a new empty topology. With the new interface lay-out (which I really like btw!) this copying is not possible any more.

grdw:
On hold after #1515 is merged.

PR #1515, opened 2016-10-21
Adding initial templates

grdw:
Pull request to extract topology templates and market model templates from used market models and topologies from unqiue LES's. A market model or topology in this setup is a clone from a template.

```
LES -> Topology     -> Topology Template
    -> Market Model -> Market Model Template
```

This PR is done for now.

Before merging; make backup of staging database.
---

Todo:
- [x] Validate migrations on exported staging database
- [x] Set topology template as featured
- [x] Clean up templates
- [x] Before merging; make backup of staging database


db/migrate/20161020101527_move_testing_grounds_topologies.rb
```diff
@@ -0,0 +1,5 @@
+class MoveTestingGroundsTopologies < ActiveRecord::Migration
+  def change
+    add_reference :topologies, :testing_ground, index: true, uniq: true, after: :graph
```
antw: `uniq` → `unique`

db/migrate/20161020090506_move_topologies_to_templates.rb
```diff
@@ -0,0 +1,24 @@
+class MoveTopologiesToTemplates < ActiveRecord::Migration
+  def change
```
antw: Minor nitpick: since most of these migrations are not reversible, `def change` should be `def up`, and there should be a `down` which raises an IrreversibleMigration:

```ruby
def up
  # ...
end

def down
  raise ActiveRecord::IrreversibleMigration
end
```

(`add_column`, `rename_column`, `create_table`, etc, are reversible; changes to data often isn't).

antw:
I think theres an issue with the migratiosn. I took a copy of the staging DB, migrated it, and a lot of the market models reference templates (via `market_model_template_id`) which don't exist.

For example, a lot of the recent market models refer to a template with ID=325, but there is no such template; there *is* a 324 and 326, but no 325.

```sql
SELECT id, market_model_template_id
FROM market_models
WHERE market_model_template_id NOT IN (SELECT id FROM market_model_templates)

# 116 rows
```

This is also true of topologies:

```ruby
SELECT id, topology_template_id
FROM topologies
WHERE topology_template_id NOT IN (SELECT id FROM topology_templates)

# 101 rows
```

I suppose it is acceptable for the market models and topologies to reference templates which no longer exist (the owner may have deleted them), but the views need to account for this. For example:

```haml
- if @testing_ground.market_model
  %span.sep
    = link_to @testing_ground.market_model.market_model_template.name, market_model_templates_path(@testing_ground.market_model)
```

This will raise `undefined method `name' for nil:NilClass` because the market model exists, but the template does not. Perhaps MarketModel should have delegation methods to deal with this?

```ruby
class MarketModel < ActiveRecord::Base
  delegate :name, to: :market_model_template, allow_nil: true
end
```

app/controllers/topology_templates_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      end,
+      filename: "#{ @topology_template.filename }.png",
+      content_type: 'image/png')
+  end
+
+  private
+
+  def topology_template_params
+    params.require(:topology_template).permit(
+      :name, :user_id, :graph, :public, :featured
```
antw: `:featured` should only be a permitted parameter for admins (this is true for market model templates also). Pundit can take care of this for you: https://github.com/elabs/pundit#strong-parameters

grdw: That's a pretty neat feature 👍 didn't know that.

spec/controllers/topologies_controller_spec.rb
```diff
@@ -2,275 +2,144 @@
...
+  #  describe "with no topology" do
+  #    let(:graph){ "" }
+
+  #    it 'does not create a 2nd topology due to errors' do
+  #      perform_clone
+
+  #      expect(Topology.count).to eq(1)
+  #    end
+  #  end
+  #end
```
antw: Can this be deleted?

spec/services/testing_ground/cloner_spec.rb
```diff
@@ -1,53 +1,52 @@
 require 'rails_helper'
 
 RSpec.describe TestingGround::Cloner do
```
antw: Did you mean to comment all this out? The Cloner class still exists.

(I just realised I've been submitting single comments instead of doing a review. Sorry for the spam. 😳)

spec/controllers/testing_grounds/caching_spec.rb
```diff
@@ -43,10 +44,10 @@
     let(:opts) { { strategies: {} } }
 
     it 'caches the data request' do
-      get :data, id: testing_ground
+      post :data, id: testing_ground
 
       loads = network.node('CONGESTED_END_POINT_0').get(:load)
-      expect(loads.length).to eq(8760)
+      #expect(loads.length).to eq(8760)
```
antw: Did this request start returning something different? I'm curious why this request has become a POST, since fetching the load data should be idempotent and doesn't alter any state on the server; did that change?

spec/controllers/topologies_controller_spec.rb
grdw: Yes, I'm getting rid of the `Cloner` and probably going to write something different for the sake of https://github.com/quintel/etmoses/issues/1499

spec/controllers/testing_grounds/caching_spec.rb
grdw: This is something I did not intent to commit at all. I'm thinking this accidentally got in.

grdw:
> I would still expect the template_id keys to be NULL

I fixed this. When somebody removes a template, the associated `market_models` and `topologies` get 'detached' from their previously associated template. I also fixed the the submitting of `featured` with pundit policies. It took a bit of fiddling but it works fine 👍 

The empty specs are removed.

antw:
Nice work @grdw! Database backup has been made; I've merged this.

Issue #1512, opened 2016-10-21
[Airbrake] [Staging] TypeError: String can't be coerced into Float


dennisschoenmakers:
**Airbrake error:** [#9018](https://quintel.airbrake.io/projects/111227/groups/1801361367394399018)
**Airbrake project:** ETMoses

**Error type:** `TypeError`
**Error message:** `` TypeError: String can't be coerced into Float ``
**Where:** `DelayedJob Worker`
**Occurred at:** Oct 21, 2016 09:43:24 UTC
**First seen at:** Oct 21, 2016 09:43:24 UTC
**Occurrences:** 0 (0 since last deploy on Oct 20, 2016 08:56:46 UTC)

**URL:** `<no information>`
**File:** `[PROJECT_ROOT]/app/models/business_case_costs.rb`

**Backtrace:**
```
[PROJECT_ROOT]/app/models/business_case_costs.rb:9:in /
[PROJECT_ROOT]/app/models/business_case_costs.rb:9:in depreciation_costs
[PROJECT_ROOT]/app/models/business_case_costs.rb:3:in total_yearly_costs
[PROJECT_ROOT]/app/models/market/initial_costs/costs.rb:17:in block in calculate
```

grdw:
Looks like someone turned `technical_lifetime` into a string in the `gas_asset_list`.

One moment please.

dennisschoenmakers:
*Comment from Airbrake:* Resolved [error](https://quintel.airbrake.io/projects/111227/groups/1801361367394399018) reoccurred.

PR #1517, opened 2016-10-21
Adding the correct number type when parsing data

grdw:
Fixes: #1512
Fixes: #1513 
Fixes: #1517 

It appeared to me that when you blank the `technical_lifetime` the `technical_lifetime` got saved to the database as `''` and not as ` `.

### DIFF
--- a/app/assets/javascripts/tables/base/editable_table.js
+++ b/app/assets/javascripts/tables/base/editable_table.js
@@ -33,10 +33,9 @@ var EditableTable = (function () {
 
     function extractTextfromCells(row) {
         return $(row).find("td.editable").toArray().map(function (cell) {
-            var input = $(cell).find("select:visible").length > 0 ? 'select' : 'input',
-                elem  = $(cell).find(input);
+            var input = $(cell).find("select:visible").length > 0 ? 'select' : 'input';
 
-            return $.trim($(cell).find(input).rawValue());
+            return  $(cell).find(input).rawValue();
         });
     }
 
--- a/app/assets/javascripts/tables/base/editable_table.js
+++ b/app/assets/javascripts/tables/base/editable_table.js
@@ -101,6 +101,7 @@ var EditableTable = (function () {
             this.rowAddedListener(clonedRow);
         }
 
+        this.tab.markAsEditing();
         this.changeListener();
     }
 
@@ -132,6 +133,7 @@ var EditableTable = (function () {
             $(this.selector).find("thead").addClass("hidden");
         }
 
+        this.tab.markAsEditing();
         this.changeListener();
     }
 
@@ -140,15 +142,6 @@ var EditableTable = (function () {
             .find("a.remove-row")
             .off('click')
             .on('click', removeRow.bind(this));
-    }
-
-    function markAsEditable() {
-        var pane = $(this.selector).parents(".tab-pane"),
-            form = pane.find("form");
-
-        $("ul.nav li a[href=#" + pane.attr("id") + "]")
-            .add(form)
-            .addClass("editing");
     }
 
     function enableLastRow(e) {
@@ -183,7 +176,9 @@ var EditableTable = (function () {
 
             $(this.selector)
                 .off('change.editable')
-                .on('change.editable', markAsEditable.bind(this));
+                .on('change.editable', function () {
+                    this.tab.markAsEditing();
+                }.bind(this));
 
             this.emptyButton
                 .off("click")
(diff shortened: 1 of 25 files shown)


app/assets/javascripts/tables/base/editable_table.js
```diff
@@ -34,9 +34,17 @@ var EditableTable = (function () {
...
-                elem  = $(cell).find(input);
+                elem  = $(cell).find(input),
+                type  = elem.data('type'),
+                value = $.trim(elem.rawValue());
+
+            if (type === 'float') {
+                value = parseFloat(value);
+            } else if (type == 'integer') {
+                value = parseInt(value);
+            }
```
antw: [The topology form uses `type=number`](https://github.com/quintel/etmoses/blob/2c874f951c0903e6d305859c85ca547d4ec05f0b/app/assets/javascripts/topology/form.js#L20-L22), but here we're using `float/integer`. Is it perhaps worth making these names consistent?

I think it would be safe to use `type === 'number' / parseFloat` even for those fields which expect an integer, since `InstalledTechnology` will convert those values to integers anyway. Alternatively if you do want to distinguish between fields which may have a fraction and those which may not, then maybe the topology form should also use `type=float`?

grdw: > Is it perhaps worth making these names consistent?

Yes. I am not sure which way or the other seems 'better'. I am not a big fan of the `number_field_tag` mainly due to it's rather crappy implementation in Firefox 😆 I'm thinking to use the `float` and `integer` fields on the topology form aswell for the sake of consistency. I'll also take a look at the `profile_matrix` because I think I skipped that one.

antw:
It seems that numeric values are being saved to the asset list as strings. I'm guessing this wasn't intentional?

**Before:**

```yaml
---
- pressure_level_index: 0
  part: pipes
  type: pe_lp
  units: 9281.578154828
  stakeholder: system operator
  building_year: 1960
  technical_lifetime: 45
  initial_investment: 130
  yearly_o_and_m_cost: 1
```

**After:**

```yaml
---
- pressure_level_index: '0'
  part: pipes
  type: pe_lp
  units: '9281.578154828'
  stakeholder: system operator
  building_year: '1960'
  technical_lifetime: '45'
  initial_investment: '130'
  yearly_o_and_m_cost: '1.1'
```

antw:
Also, updating a LES topology is returning an error due to an `units` input being blank:

* Graph may not have a non-numeric "units" attribute.

I wouldn't object to defaulting topology nodes to have an explicit `units` value of 1, and showing a validation error if the user deletes the value. However I think this change in behaviour also wan't intentional?

grdw:
>  I'm guessing this wasn't intentional?

No, it's even a bit odd looking at it. 🤔  I explicitly parse the numbers to floats or integers and than save the result. I think I might know what this could be though.

> I wouldn't object to defaulting topology nodes to have an explicit units value of 1, and showing a validation error if the user deletes the value. However I think this change in behaviour also wan't intentional?

No, this wasn't intentional aswell. The thing is that `units` probably got saved as `''` being parsed as 0 because `''.to_i == 0`. I'll take a look at this aswell.

grdw:
@antw 

> It seems that numeric values are being saved to the asset list as strings.

Turns out `$.trim` turns an integer or float back again to a string 😅 

> Graph may not have a non-numeric "units" attribute.

This was because I changed the from `data('type')` to `data('dataType)` and forgot to rename the topology form 😶

grdw:
@antw if you could take a look.

antw:
`pressure_level_index` is still being cast to a string in the DB when previously it was an int, but otherwise everything else looks good.

antw:
Ahh... I spoke too soon. Strings with spaces in them – such as "system operator" and "heat producer" stakeholders – seem to have the spaces removed:

```yaml
# Heat source list stakeholder:

- key: central_heat_network_dispatchable
  heat_capacity: 10.660117200536018
  units: 1
  stakeholder: heatproducer # <-----
  total_initial_investment: 100
  technical_lifetime: 30
  om_costs_per_year: 2
  marginal_heat_costs: 0.022
  distance: 1
  priority: '1'

# Gas asset list stakeholder

---
- pressure_level_index: '0'
  part: pipes
  type: pe_lp
  units: 9281.578154829
  stakeholder: systemoperator # <-----
  building_year: 1960
  technical_lifetime: 45
  initial_investment: 130
  yearly_o_and_m_cost: 1.1
```

grdw:
> stakeholder: systemoperator # <-----

This was due to [`[\'\" ]`](https://github.com/quintel/etmoses/pull/1517/files#diff-d6a615214726b3886650fcf48314e4fdR71) instead of the correct one: `[\"\']`. I don't know how that extra space got in there.

The `pressure_level_index` missed a type, I checked and the `priority` of the dispatchable heat source table were also saved as strings which should have been integers, and the `demand` of the `technology_profile_matrix` was an `integer` which should have been a `float`.

I also noticed that `dispatchable` is being saved as en empty string with must run heat sources but I'm not sure why this is again. 🤔

antw:
This looks good. :+1:

@grdw Could you look into the merge conflict when you have a moment? The `master` branch has changes to `save_all.js: function change(e)`, but this PR removes that function entirely.

As soon as that's taken care of this can be merged (you're welcome to do it yourself if you wish).

grdw:
> you're welcome to do it yourself if you wish

I had to make a minor change to the technologies form to detect that someone selected a new `csv` but that's been taken care off. I'll merge this personally.

PR #1521, opened 2016-10-24
User may customise heat capacity per connection

antw:
This PR adds an input on the testing ground "general" form allowing the visitor to customise the central heat buffer volume per heat-connection. Defaults to `nil/NULL`, which in turn will default to 10kWh in `Network::Builders::Heat` (so that if we change the default in the future, older LESes won't be stuck with an out-of-date value).

screen shot 2016-10-24 at 14 25 50 screen shot 2016-10-24 at 14 25 59

I think this field would be more at home on the "Heat source list" tab, but I'm not sure if that's possible? All the fields on that page relate to the heat source list, while this is a column on the `testing_grounds` table. I think this is not a deal-breaker.

@grdw Can you take a look at this when you have the time? Thanks!

Ping @ChaelKruip

### DIFF
--- a/app/models/network/builders/heat.rb
+++ b/app/models/network/builders/heat.rb
@@ -35,7 +35,11 @@ module Network
         @park = Network::Heat::ProductionPark.new(
           must_run:         must_run,
           dispatchable:     dispatchable,
-          # https://github.com/quintel/etmoses/issues/971
+          # Normal heat producers may fill the buffer up to base_volume (by
+          # default, 10kWh), but must-runs may add more to the buffer so as not
+          # to waste excess energy (17.8kWh assuming base volume of 10kWh).
+          #
+          # See https://github.com/quintel/etmoses/issues/971
           volume:           base_volume,
           amplified_volume: base_volume * 1.78
         )
--- a/app/models/network/builders/heat.rb
+++ b/app/models/network/builders/heat.rb
@@ -3,10 +3,12 @@ require_relative '../builders'
 module Network
   module Builders
     class Heat
-      def self.build(tree, techs = TechnologyList.new, heat_sources = nil)
+      def self.build(tree,techs = TechnologyList.new, sources = nil, opts = {})
         new(
-          tree, techs,
-          heat_sources ? HeatSourceListDecorator.new(heat_sources).decorate : []
+          tree,
+          techs,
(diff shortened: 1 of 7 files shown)


app/models/network/builders/heat.rb
```diff
@@ -16,23 +18,26 @@ def to_graph
...
+        base_volume = number_of_connections * 4 * @capacity_per_conn
+
         @park = Network::Heat::ProductionPark.new(
           must_run:         must_run,
           dispatchable:     dispatchable,
           # https://github.com/quintel/etmoses/issues/971
-          volume:           number_of_connections * 4 * 10.0,
-          amplified_volume: number_of_connections * 4 * 17.78
+          volume:           base_volume,
+          amplified_volume: base_volume * 1.78
```
grdw: `1.78` is a bit of magic number to me. What does this mean in this context?

antw: I've expanded on the comment with information about the number; I agree that the comment should have just described it in the first place, instead of directing the reader to a GitHub issue. :+1:

grdw:
I have no idea how to properly test this. It doesn't cause any errors when changing the number, saving works, all the basics still function + the specs succeed 👍 . I don't see any difference in my heat calculation but that can be because I don't know exactly what to look for.

grdw:
> I think this field would be more at home on the "Heat source list" tab, but I'm not sure if that's possible?

It could be. If you edit the 'testing_grounds/edit.html.haml' and add another `form` to the pane it will probably work 👍.

antw:
> I have no idea how to properly test this.

We don't have specs for the network calculation, so you'll have to take my word for it. 😆  Here are some screenshots from a local LES:

### 0kWh per connection (no buffering occurs):

screen shot 2016-10-24 at 16 53 26

### No value set (defaults to 10kWh):

screen shot 2016-10-24 at 16 53 14

### 100kWh per connection:

screen shot 2016-10-24 at 16 52 59

grdw:
> so you'll have to take my word for it.

````ruby
expect(true).to eq(true)
````

antw:
> expect(true).to eq(true)

[So about that...](https://github.com/quintel/etmoses/pull/1521/files#diff-2ead4cd752449974b534cd433c354a99L46) 😱

Issue #1528, opened 2016-10-26
Split between individual and aggregated load profiles is not always practical


DorinevanderVlies:
### Current situation:
Currently for <31 units an individual load profile for electricity has to be selected.
For >30 households an aggregated profile has to be selected.

### Problem:
There are cases when I want to assign an aggregated profile to <31 households.
For example when I have different end nodes to represent different stakeholders. But, on the network these nodes are connected to one transformer. In this case the individual profiles are way too spiky.

### Possible solutions:
- Add a button which can be selected if a user wants to use an aggregated profile on <31 households. 
- Keep the current interface, but make it possible to assign an aggregated profiles on <31 households when a CSV file is used to define the technologies.

ChaelKruip:
> Keep the current interface, but make it possible to assign an aggregated profiles on <31 households when a CSV file is used to define the technologies.

I prefer this option. 

@antw @grdw what do you think?

ChaelKruip:
@grdw can you pick this up?

DorinevanderVlies:
An other option is that the difference between individual and aggregated profiles is just defined in the name of the profile.  So, not all profiles can be used for all numbers of households. Like it is for water heating.

I think I prefer this option.

grdw:
We could make this way more easy. The easiest way would be getting rid of the `< 31` threshold and make it possible to create a separate `base_load_edsn` and `base_load` technology. I think it's a bit 'unfair' to only make this possible with a CSV input (in fact I'm not a big fan of the CSV input) and not in the profile matrix editor.

antw:
> I think it's a bit 'unfair' to only make this possible with a CSV input (in fact I'm not a big fan of the CSV input) and not in the profile matrix editor.

I'm not sure of the best solution, but I absolutely agree with this. We can't apply one set of rules to the normal interface and another to CSVs, otherwise things will break in unpredictable ways when you try to edit these LESes.

I, too, wish we could get rid of the CSV import (export is fine). It might be useful to know in which ways editing a CSV is easier than the normal technology editor, and try to improve it until the CSV import feature is no longer needed. That's a bit off-topic though...

antw:
> There are cases when I want to assign an aggregated profile to <31 households. For example when I have different end nodes to represent different stakeholders. But, on the network these nodes are connected to one transformer. In this case the individual profiles are way too spiky.

[...]

Issue #1459, opened 2016-09-28
Dragging the topology editor causes the topology to jump into the upper left corner


grdw:
This only happens the first time you do this. Needs some investigating.

PR #1539, opened 2016-10-31
Allow to freely add base load edsn techologies

grdw:
Fixes: #1528

### DIFF
--- a/app/assets/javascripts/les/profile_matrix_editor/edsn_switch.js
+++ b/app/assets/javascripts/les/profile_matrix_editor/edsn_switch.js
@@ -29,11 +29,9 @@ var EdsnSwitch = (function () {
 
     return {
         cloneAndAppendProfileSelect: function (target) {
-            var target     = $(target),
-                type       = target.data('type'),
-                isImported = target.hasClass('imported');
+            var type = $(target).data('type');
 
-            if (validBaseLoads.test(type) && isImported) {
+            if (validBaseLoads.test(type)) {
                 swapSelectBox.call(target);
             }
         }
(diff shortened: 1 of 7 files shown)


antw:
This looks fine, apart from one small issue.

I'm using a copy of the staging database locally, and started to import scenario 596307 from beta ETEngine. On the technology import form (prior to actually saving the new LES), the "Households aggregated" techs were assigned the "Ameland tourist profile electricity" profile in spite of that profile being excluded from auto-assignment:

--

screen shot 2016-11-02 at 09 26 50

--

screen shot 2016-11-02 at 09 29 17

antw:
> Sort profiles on `included in concurrency`

That feels like an awkward solution. Isn't there a way to add `selected="selected"` to the first `<option>` which is permitted to be auto-selected? Now "Ameland" appears beneath "Base-load" which isn't intuitive:

screen shot 2016-11-02 at 12 10 03

antw:
> > Sort profiles on `included in concurrency`
>
> That feels like an awkward solution.

Perhaps a nice compromise is to do something like #1459 and sort profiles into two optgroups, with `included_in_concurrency` being called "Featured", and everything else called "Other"?

That has the advantage of making it obvious to a visitor which profiles are official, and which have been uploaded by third parties?

```
+--------------------------------------+---+
| Profile                              | v |
+--------------------------------------+---+
| Featured                                 |
|   Base-load profile EDSN type E1A        |
|   Base-load profile EDSN type E1B        |
|   ...                                    |
| Others                                   |
|   Ameland tourist profile electricity    |
|   ...                                    |
+------------------------------------------+
```

PR #1540, opened 2016-10-31
Clean up old concurrency code

grdw:
When working on #1539 I found a lot of old concurrency code which wasn't used anymore. I cleaned it.


antw:
Thanks @grdw, nice catch! The best code is deleted code! :grinning:

I merged this manually to fix a conflict.

PR #1548, opened 2016-11-02
Add 'save as' topology template feature

grdw:
Fixes: #1499


antw:
screen shot 2016-11-03 at 10 18 48

screen shot 2016-11-03 at 10 19 01

Weirdly both translations can be found in the locale files, and I18n seems to have been configured to include them in the JS. Can you reproduce this or is it just me?

[The validation of the name](https://github.com/quintel/etmoses/pull/1493) seems to have been lost (but this is also true on beta). A `422 Unprocessable Entity` request comes back to the browser, but no message is displayed. Since this is also true on beta – and therefore not related to this PR – we can merge this provided the translations work for you, and fix the validation separately.

grdw:
> Weirdly both translations can be found in the locale files, and I18n seems to have been configured to include them in the JS. Can you reproduce this or is it just me?

You need to, stop the server, do `rake tmp:clear`, start the server (this took me some time to figure out once)

antw:
> You need to [...] `rake tmp:clear`

TY! Works perfectly. :+1:

Issue #1547, opened 2016-11-02
Creating a new topology template gives error message


DorinevanderVlies:
I try to create a new [topology template](https://beta-moses.energytransitionmodel.com/topology_templates/422). But it does not save properly and editing it results in an error. Below some screenshots.

### Creating a topology
![image](https://cloud.githubusercontent.com/assets/19907658/19922941/354f98d8-a0e5-11e6-931a-edd650a2888c.png)

### After saving it
![image](https://cloud.githubusercontent.com/assets/19907658/19922953/4464e24c-a0e5-11e6-9bc0-73386429def3.png)

### After pressing edit
![image](https://cloud.githubusercontent.com/assets/19907658/19922965/59205cd4-a0e5-11e6-803a-4dca5dda7f86.png)

ChaelKruip:
Can you pick this up with priority @grdw?

grdw:
Yes I can 👍

DorinevanderVlies:
Issue solved

PR #1550, opened 2016-11-02
Initialize topology template with default template

grdw:
Also fixes authentication bug #1547


antw:
This is much nicer than starting with just one node. :+1:

Issue #1562, opened 2016-11-11
Heat assets keep changing back to default settings


DorinevanderVlies:
I have a couple of LESses in which I want to set investment and O&M costs for the heat assets.
However, they keep changing back to default settings. Sometimes my changes hold for a couple of business case runs. But for example this morning I looked at my LES again and the setting were default again.

[Example of LES.](https://beta-moses.energytransitionmodel.com/testing_grounds/809)
These are the setting I want to use, and have frequently filled in.
![image](https://cloud.githubusercontent.com/assets/19907658/20211387/981131c8-a7fe-11e6-9190-284ec015e208.png)

Keeps changing back to this:
![image](https://cloud.githubusercontent.com/assets/19907658/20211321/3b1fc68c-a7fe-11e6-811c-9a20a04482d0.png)

notifying @AlexanderWirtz

DorinevanderVlies:
It seems that every time a small change is made to the heat sources the investment and O&M costs of the heat assets go back to their default settings.

antw:
[It looks like this is intentional](https://github.com/quintel/etmoses/blob/d2e923874f670a01a09f65b849fa7d3c68afe2d2/app/controllers/heat_source_lists_controller.rb#L9), but I don't know why. @grdw can you shed any light on this?

grdw:
If you alter the heat sources, the primary heat asset list is regenerated from scratch. The reason for that is something I'm not quiet sure of aswell 😆  

But to reason it the other way around; how would we know which primary heat asset belongs to which heat source? Right now there's no foundation in place to do this properly. The correct way to do this is to give every newly added `heat_source` an unique id. Than when generating the `heat_asset_list`, associate the two. Than when updating, ... etc. 😄

DorinevanderVlies:
> how would we know which primary heat asset belongs to which heat source?  

The heat assets are associated with a source. However, when two sources of the same type are present (with the same distance) it is indeed not defined which assets belongs to which source. 

A possible solution would be to have a choice between "save" and "save and update heat assets". However, I'm not really in favour of this option myself, as I think it will lead to confusion.

> The correct way to do this is to give every newly added heat_source an unique id. Than when generating the heat_asset_list, associate the two. Than when updating, ... etc. 😄

Sounds like a better solution than mine

grdw:
[...]

PR #1577, opened 2016-11-29
Manually reload heat asset list

grdw:
Fixes: #1562


antw:
Looks good in that the heat asset attributes no longer get overwritten by changing the sources.

For me the "Rescale" button isn't updating the heat asset list table. The backend action is completing and refreshing the page *does* show the new values. The returned JS does seem to include a line to do this (`$(".primary-table").html(...)`), but doesn't have any effect for me; I think the JavaScript isn't being executed?

grdw:
The code did got executed but there was a intermediary form submit that caused the code after that submit not to be executed (I have no idea why, probably because you bounce of the page somewhow?). I edited it and it should now work.

Issue #1586, opened 2016-12-14
Secondary heat infrastructure costs not represented in business case


DorinevanderVlies:
### Description

This [LES](https://beta-moses.energytransitionmodel.com/testing_grounds/862/edit) only has technologies attached to node "1 - flat". I just added heat technologies (all district heating) to this node. No investment and O&M costs attached to these technologies.

In the heat assets tab I assigned 1000 euro per connection to all 54 connections. The lifetime is 10 years. So I would expect (1000 euro/10year lifetime *54 units = )5400 euro depreciation costs of stakeholder 1. However, there are no depreciation costs for stakeholder 1.

### Screenshots
![image](https://cloud.githubusercontent.com/assets/19907658/21180945/ad5d3c36-c1fb-11e6-9651-d5991755794e.png)

![image](https://cloud.githubusercontent.com/assets/19907658/21180810/f5ec591a-c1fa-11e6-80b4-9ef80ee87e28.png)

notifying @AlexanderWirtz

grdw:
I'll have a look.

grdw:
If I understand this issue correctly the `Customer Ac1` should receive depreciation costs?

DorinevanderVlies:
Yes, indeed. Or extra outgoing costs, but this doesn't seem to be the case as well.

grdw:
I see what's going on here. Appearantly the `number_of_units` is always 0 and I also know why. Am fixing it.

grdw:
Ok, that was one. But there's another thing going on here. The topology also contains the `Customer AC1` but that calculation returns `NaN` for a reason. Probably because the `technical_lifetime` is not set? If you could check hat out.

DorinevanderVlies:
I found technical lifetimes and investment costs defined as 0 for the buffer. This could be a problem. It shoudn't really be a problem right? 

Are the two LESses below working properly?

[LES with technical lifetimes and initial investment and O&M left emtpy](https://beta-moses.energytransitionmodel.com/testing_grounds/862)

[LES with technical lifetime of 30 years even when initial investment and O&M costs are non-defined (empty)](https://beta-moses.energytransitionmodel.com/testing_grounds/863)

grdw:
> I found technical lifetimes and investment costs defined as 0 for the buffer. This could be a problem. It shoudn't really be a problem right?

[...]

PR #1587, opened 2016-12-14
Setting the correct number of units for secondary heat assets

grdw:
Fixes: #1586

### DIFF
--- a/app/models/market/from_market_model_builder.rb
+++ b/app/models/market/from_market_model_builder.rb
@@ -16,7 +16,7 @@ module Market
     end
 
     def to_market
-      Builder.new(data).to_market
+      Market::Builder.new(data).to_market
     end
 
     private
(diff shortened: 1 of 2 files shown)


antw:
Is it possible to instead determine the number of units during the cost calculation? I thought we avoided setting a NoU attribute on heat assets since the number of heat connections could change when the user added or removed "district heat" technologies, and this wouldn't be reflected in the assets when using an explicit NoU?

grdw:
> Is it possible to instead determine the number of units during the cost calculation? 

That's correct. I'll see what I can manage.

grdw:
@antw I mange to include the `heat_connections` in the initial market calculation for heat assets. 👍 So no noh in the db.

antw:
Looks good! :+1:

Issue #1590, opened 2017-01-06
Capping of solar PV not taken into account in stacked chart


DorinevanderVlies:
### Summary
Capping of solar PV is shown in the total chart, but not in the stacked chart. (the strategies switch is activated)

The third week June of [_Node 3 - portiekflat_ of this LES](https://beta-moses.energytransitionmodel.com/testing_grounds/894) shows a difference with and without strategies in the total chart. However, the stacked chart is the same with and without strategies.

For example for EV strategies I did see differences in the stacked chart with and without strategies.

### Images
![image](https://cloud.githubusercontent.com/assets/19907658/21713907/95779612-d3fc-11e6-94b6-8acd2af20503.png)
![image](https://cloud.githubusercontent.com/assets/19907658/21713912/a61f294e-d3fc-11e6-9b27-fe382fe783e3.png)
![image](https://cloud.githubusercontent.com/assets/19907658/21713919/ac4dafe8-d3fc-11e6-9cb2-5b429f3cfebb.png)

ChaelKruip:
@antw @grdw do you have any ideas why this is?

grdw:
My gut feeling tells me that this is probably a JavaScript bug. @DorinevanderVlies  Could you try it with another strategy and see if the changes are reflected in the stacked chart?

DorinevanderVlies:
I made the [LES even simpler](https://beta-moses.energytransitionmodel.com/testing_grounds/894).

> Could you try it with another strategy and see if the changes are reflected in the stacked chart?

Other strategies in this LES also cause trouble when shown in the stacked chart. In this example I selected "storage in batteries prioritising local PV production". The nodes are serial. However node K+Z 2 shows a different stacked chart than node 2 - portiekflat.

The total chart is, as it should be, the same for both nodes.

![image](https://cloud.githubusercontent.com/assets/19907658/21716053/57adc228-d408-11e6-9d54-8a610dc2741c.png)

![image](https://cloud.githubusercontent.com/assets/19907658/21716069/6a85587a-d408-11e6-8776-7d063fad0143.png)

grdw:
I don't really know what you're saying (sorry), but is the shape of the stacked chart (with strategies) the same as the full/non-stacked chart without strategies for a single node?

PR #1592, opened 2017-01-06
Consume conserved PV loads using sub-path

antw:
Fixes that conserved PV production was not removed from the tech-load (#1590). Sadly there are no tests for this part of the calculation (:disappointed:), but I have checked manually:

| Before | After |
| :---: | :---: |
| screen shot 2017-01-06 at 11 56 13 | screen shot 2017-01-06 at 11 52 22 |


grdw:
Just tested this locally and it seems to be working 👍 I'll merge this.

PR #1612, opened 2018-02-02
Add technology uglifier for LES's

grdw:
This seems to work locally:

screen shot 2018-02-02 at 15 40 26

However, if I try to save it (it does save):

Something goes wrong at ETEngine (for `api/v3/scenarios/<scenario-id>/converters/stats`):

screen shot 2018-02-02 at 15 41 01

Seems that strong parameters is needed for this to work. 🤔

### DIFF
--- a/app/models/installed_technology.rb
+++ b/app/models/installed_technology.rb
@@ -286,7 +286,10 @@ class InstalledTechnology
   end
 
   def to_h
-    super.delete_if { |_, v| v.blank? }
+    tech = super.delete_if { |_, v| v.blank? }
+    tech[:associates] = (tech[:associates] || []).map(&:to_h)
+    tech[:components] = (tech[:components] || []).map(&:to_h)
+    tech
   end
 
   private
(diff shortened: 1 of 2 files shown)


antw:
Sorry for being thick: how does converting technologies to a hash fix the 401 Unauthorized response? Was ETMoses sending data in the wrong format?

grdw:
Oh, I'm sorry for not explaining.

This is the current sitaution. Currently on [beta](https://beta-moses.energytransitionmodel.com/testing_grounds/import) there's still an error ocurring when trying to import a scenario for ETMoses. This is the [error](https://quintel.airbrake.io/projects/111227/groups/2141423638789206726?tab=overview). The problem is the formatting.

After the couple of lines of fixing in this PR the other error bubbles up at ETEngine (probably due to the Rails 5 upgrade I can imagine).

antw:
Got it, thanks!

You're right about the strong params in ETEngine. ConverterController#stats needs the following change:

```diff
- keys = params.require(:keys)
+ keys = params.require(:keys).permit!.to_h
```

grdw:
I created a PR here: https://github.com/quintel/etengine/pull/999

PR #1616, opened 2018-02-15
Issues/#1613 2

grdw:


antw:
This looks fine, but in hindsight I'm a bit confused about the `storage.volume`/`storage_volume` thing. Since the `storage` attribute was already whitelisted in ETEngine, is that change actually needed? If so, do you know why?

grdw:
I see that `storage` is already whitelisted. But it's volume specifically isn't. When I remove the `storage_volume` whitelisting the following happens:

screen shot 2018-02-16 at 09 06 32

Maybe if I replaced `storage_volume` with just `storage` in the technology importable attributes and let the `StorageVolumeAttribute` figure out how to fetch the `volume` from the `storage` hash? Maybe that will make more sense.

One sec.

app/models/import/technologies/fetcher.rb
```diff
@@ -33,10 +33,11 @@ def response
       end
 
       def keys
-        #Hash[technologies.map do |technology|
-        #  [technology.key, technology.importable_attributes]
-        #end]
-        Technology.importable.map(&:key)
+        Hash[Technology.importable.map do |technology|
+          [technology.key, technology.importable_attributes.map do |key|
+            key.sub(/\..+/, '')
```
grdw: @antw this also seems to fix the problem 😆 but this feels a bit like a hack.

antw: > @antw this also seems to fix the problem

I think what's been going on is that previously ETMoses was just sending the array of technology keys to ETEngine:

```json
{ "keys": ["tech_1", "tech_2", "...", "tech_n"] }
```

... instead of a hash of technologies and attributes:

```json
{
  "keys": {
    "tech_1": ["attr_1", "attr_2"],
    "tech_2": ["attr_1", "attr_3"],
  }
}
```

... and as luck would have it, ETEngine was happy with that and would reply with *all* the data for the requested technologies.

I'm guessing this stopped working when `permitted_params` was changed [to require `:keys` be a hash](https://github.com/quintel/etengine/commit/10292733c9600f68fe4c275d69ef69b896464f73).

> this feels a bit like a hack.

I agree, and I think I know why. This change means that the fetcher is now not only responsible for *fetching*, but also about the internal details of the keys that need to be fetched: it needs to know that some keys have a `.` and to ignore everything including and after that character.

It also feels hacky because there are already classes responsible for attributes: `NumberOfUnitsAttribute`, `StorageVolumeAttribute`, etc.

I've tried to resolve this on [the "issues/#1613-3" branch](https://github.com/quintel/etmoses/tree/issues/%231613-3) which:

* Adds `remote_key` to Attribute – make the attribute responsible for knowing how the attribute maps to ETEngine: it already has `remote_name` so I don't *think* this is too much responsibility for Attribute.

* Adds `importable_attributes` method to TechnologyBuilder: returns an enumerable of Attribute objects for the technology. Changed `build` so it also delegates to `importable_attributes`.

* Call `TechnologyBuilder.importable_attributes` from the fetcher.

What do you think? I can open a PR if you think it's worth proceeding with it.

grdw: > What do you think? I can open a PR if you think it's worth proceeding with it.

This seems like the right type of solution but it introduces quite some technical overhead for a single attribute imo. I'd rather see the API of ETEngine only knowing the `storage_volume` attribute (while removing the `storage` attribute from the whitelisting). 

So in https://github.com/quintel/etengine/blob/master/app/models/api/v3/converter_stats_presenter.rb : 

```ruby
DEFAULTS = {
  # remove storage: nil
}

ATTRIBUTES = {
  ..,
  storage_volume: -> (conv) { conv.storage.volume },
  ..
}
```

And whitelist `storage_volume` in the technologies. I don't feel that ETMoses should be the one to know that one attribute returns a `Hash` and the other a `Float`.

antw:
Just gonna merge this PR for now.

> I don't feel that ETMoses should be the one to know that one attribute returns a Hash and the other a Float

FWIW I disagree with this. I don't think ETEngine's public API should change to accommodate a (mostly) dead app like ETMoses. It returns an object/hash for `storage` because other users of the API might be interested in the other storage keys (`decay` and `cost_per_mwh`). Revoking the whitelisting of `storage` would break those clients.

PR #1617, opened 2018-04-03
Fix fetching stats for a single converter

antw:
Fixes `EtEngineConnector` to request all converter attributes when `keys` is an array. Fixes the technology selector which currently fails to fetch the attributes:

screen shot 2018-04-03 at 20 39 11

Ref: [Airbrake](https://quintel.airbrake.io/projects/35333/groups/2184964795887801140?tab=overview)

### DIFF
--- a/lib/et_engine_connector.rb
+++ b/lib/et_engine_connector.rb
@@ -20,8 +20,17 @@ class EtEngineConnector
 
   def stats(scenario_id)
     url = ETM_URLS[:stats] % [@provider, scenario_id]
+    params = @params
 
-    et_api_request(:post, url, @params, HEADERS)
+    if params.key?(:keys) && params[:keys].is_a?(Array)
+      # 'stats' wants a hash of keys and attributes to fetch. If the caller does
+      # not care about specific attributes, request all.
+      #
+      # { keys: [:a, :b] } becomes { keys: { a: nil, b: nil } }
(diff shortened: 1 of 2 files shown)


lib/et_engine_connector.rb
```diff
@@ -20,8 +20,17 @@ def initialize(params = {})
...
   def stats(scenario_id)
     url = ETM_URLS[:stats] % [@provider, scenario_id]
+    params = @params
+
+    if params.key?(:keys) && params[:keys].is_a?(Array)
+      # 'stats' wants a hash of keys and attributes to fetch. If the caller does
+      # not care about specific attributes, request all.
+      #
+      # { keys: [:a, :b] } becomes { keys: { a: nil, b: nil } }
+      params = params.merge(keys: Hash[params[:keys].zip([])])
```
grdw: This bit looks a bit hacky. At least - I have a hard time explaining it to my self. The comment above does explain it, but maybe it can become a bit more clear if `merge` isn't used. The `.zip([])` can be replaced by something else. Maybe do something like:

```ruby
params.store(:keys, params[:keys].each_with_object({}) do |key, obj|
  obj[key] = nil
end)
```

But I don't feel to strongly about this though.

grdw:
I'll merge this for now, considering that Airbrake is throwing errors because of this.

